PR #857, opened 2017-08-28
Update OBI build system

jamesaoverton:


jamesaoverton:
Reasoning over OBI inside Travis failed because of a 10 minute time-out.

PR #913, opened 2018-03-12
testing with test obsolete

cstoeckert:


jamesaoverton:
Hmm. Some of those tests should have failed. I'll look into it.

Thanks for testing this Chris! I'll close this now (without merging).

PR #914, opened 2018-03-19
changed derives from for human specimen set to derives from part of. …

bpeters42:
…#883


jamesaoverton:
This was an experiment, but OBI has changed enough that it would now be easier to make this change by starting over.

Issue #919, opened 2018-04-17
We have been using the wrong dublin core URI to indicate license all this time


cstoeckert:
This is based on:
https://github.com/OBOFoundry/OBOFoundry.github.io/issues/540

The OBOF 'open' principle (http://www.obofoundry.org/principles/fp-001-open.html) now recommends dcterms:license as opposed to dc:license which is what we use.

turbomam:
Agreed, Mark will change.

Somebody... add this new policy to XYZ listserv?
James and Becky will work on this globally?

zhengj2007:
license information is added during the release process. @jamesaoverton Could you please take care of it on next OBI release? Besides, we need to add: xmlns:dcterms="http://purl.org/dc/terms/". Thanks!

jamesaoverton:
It's not added during release, it's part of `obi-edit.owl` and can be edited in Protege.

turbomam:
Jie, I just opened obi/src/ontology/obi-edit.owl in Protege, to work on issues I volunteered for last week.

Does this mean that I shouldn't modify 

```
   <owl:Ontology rdf:about="http://purl.obolibrary.org/obo/obi/dev/obi-edit.owl">
        <dc:license rdf:resource="http://creativecommons.org/licenses/by/3.0/"/>
    </owl:Ontology>
```

zhengj2007:
@jamesaoverton Sorry. I did not remember correctly.

@turbomam Please go ahead to make the changes. Besides, please change 3.0 to 4.0, the latest version of CC BY.

turbomam:
@zhengj2007 
@jamesaoverton 

OK.  

I cloned the OBI github repo and started editing... now I'm thinking that I should have forked first.  Can anyone remind me where James' obi-editing/pull-requesting documentation is?

zhengj2007:
@turbomam I think you make the changes and create pull request when commit your changes. 

The instruction is on the page:
https://github.com/obi-ontology/obi
under "Make changes" section

jamesaoverton:
The README here <https://github.com/obi-ontology/obi> has a recommended editing workflow.

turbomam:
Thanks, James.  I saw your other comments, and I'll do better commenting, issue separation, and non-button pushing next time.

jamesaoverton:
Much appreciated!

Issue #859, opened 2017-08-30
Update OBI:0002112


jahilton:
When requesting a term for 'small RNA-seq', we cited a product that unfortunately, specified a size range that does not reflect the common usage of this assay name.
One of the earliest publications notes a size of <200 nt (http://science.sciencemag.org/content/316/5830/1484.full). Could you please update the definition to the broader size range?

We would also like to request an alternate term of 'short RNA-seq'
reference: http://genome.cshlp.org/content/21/2/182.full (this also references <200 nt)

turbomam:
Edited locally, haven't pull-requested yet.

Issue #831, opened 2017-06-05
NTR: nCounter Mouse miRNA Expression arrays


jahilton:
editor preferred term: nCounter Mouse V1.5 miRNA Expression array
textual definition: An array which is manufacutred by NanoString Technologies. Built upon color-coded molecular barcodes technology, the array profiles mouse miRNA with increased specificity and sensitivty than microarrays.
definition source for the textual definition: https://www.nanostring.com/products/mirna-assays/mirna-panels, https://www.nanostring.com/application/files/3114/8964/6980/PDS_Mouse_v1_5_miRNA_update.pdf
logical definition: assay array --> nCounter Mouse V1.5 miRNA Expression array

The lab has also requested OBI terms for two previous versions of the same array, but the company no longer has documentation available for those.

editor preferred term: nCounter Mouse V1.2 miRNA Expression array
logical definition: assay array --> nCounter Mouse V1.5 miRNA Expression array
Usage: http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0100951

editor preferred term: nCounter Mouse V1.3 miRNA Expression array
logical definition: assay array --> nCounter Mouse V1.5 miRNA Expression array
Usage: http://iovs.arvojournals.org/article.aspx?articleID=2319723

Please advise if we should contact Nanostring about documentation or if the references given are sufficient enough.

DanBerrios:
I don't know that "array" is the correct term to use for this class of assay.  "Microarrays" are almost always 2-dimensional.  Yes, there are 1-dimensional "arrays" in mathematics, but most call those "vectors," and in the company docs they don't refer to this as a kind of an array, they only use the terms "assay" and "panel."

cstoeckert:
@DanBerrios assay array covers more than microarrays and we do already have 'nCounter Human V2 miRNA Expression array' in OBI. 
@jahilton Given the definition of 'nCounter Human V2 miRNA Expression array' we need additional info to differentiate the different versions in the definition. An alternative way to go is to just define a 'nCounter Mouse miRNA Expression array' which is independent of version. We can define as having a 'version number' (http://purl.obolibrary.org/obo/IAO_0000129) that can be used to specify the different versions.

jahilton:
I have heard back from the manufacturer...

> There aren’t any content changes between the different versions of the Mouse miRNA panels (all of the targets stay the same).
[...]

PR #922, opened 2018-05-11
issues discussed 20180230 #919 #859 #831

turbomam:
edited but didn't run makefile


jamesaoverton:
Life is easier if you make one commit per issue fixed, rather than combining them like this. It makes it easier to understand the history and to roll-back changes if there are problems.

I would prefer to be the one who presses the "merge" button.

PR #947, opened 2018-07-30
Sequence analysis

DanBerrios:
This replaces the pull request that was incorrectly generated.  New terms template in the area of sequence analysis.


jamesaoverton:
@DanBerrios For 'sequence data feature count tabulation' the 'Input' column has 'sequence alignment'. But that's a process, and causes a reasoning error. I think the input should be 'sequence data'. Can you please confirm?

DanBerrios:
Input for sequence data feature count tabulation should be aligned sequence data

Issue #927, opened 2018-05-12
term request 'digital acquisition card'


rvita:
term request by Brenda Farrell for 'digital acquisition card'
definition: The digital acquisition card acts as the interface between the computer and a measurement device via the computer bus. It converts the current or voltage generated by a measurement device to a analog signal via a analog to digital convertor and it also generates the stimulus voltage or current that is fed to a measurement device via a digital to analog converter.
alternative term: DAQ device, digital acquisition and digital generation card
parent class: device
output (?): device, 'has function' some 'signal conversion function'
post OBI conference call 4/23/2018 for bfarrell@bcm.edu

turbomam:
"Converts signal to analog via a ADC"?  Need clarification.

gaviganet:
I see my logic was backwards.  Is this better?

> The digital acquisition card acts as the interface between the computer and a measurement device via the computer bus. It converts the signal generated by a measurement device (output from device) to a digital signal via an analog-to-digital converter, ADC. It also converts a digital signal to an analog signal via a digital-to-analog converter, DAC. This analog signal (e.g., stimulus) is fed to a measurement device (input to device)."

turbomam:
Yes, that textual definition makes more sense.  Thanks for replying.  I'll add it this week.

PS:  request originally from Brenda F, an Otolaryngologist

rvita:
adding the documents Dr Farrell showed on the OBI call regarding her term requests

[Fig1_flow_chart&legend.pdf](https://github.com/obi-ontology/obi/files/2156095/Fig1_flow_chart.legend.pdf)
![bfdiagram](https://user-images.githubusercontent.com/13909107/42176875-57cb381c-7ddf-11e8-84af-cf0f64c37e5c.jpg)

PR #949, opened 2018-08-07
DAQ card, issue #927

turbomam:
added single term 'digital acquisition card' based on issue #927
I tried to configure Protege to pick the next available OBI_nnn identifier, and it picked OBI_0000003


rctauber:
Updated to have ID OBI:0002586, the next available ID after the terms from #947.

PR #955, opened 2018-08-22
Remove spurious underscores

jamesaoverton:


jamesaoverton:
@rctauber Any idea how these underscores snuck in? Probably my fault.

rctauber:
I just did some digging and it looks like they were there when I first created the template, but I don't see them in the Google spreadsheet. I have no idea where they came from, maybe something weird with copying the Google sheet?

jamesaoverton:
Yeah. Weird.

Issue #942, opened 2018-06-08
Correct definition, alt terms for OBI_0000650 differential expression analysis data transformation


DanBerrios:
Correct defintion

FROM: 

A differential expression analysis data transformation is a data transformation that has objective differential expression analysis and that consists of

TO: 
A  differential expression analysis data transformation is a data  transformation that has objective of differential expression analysis,

Add alternative term:

differential expression analysis

zhengj2007:
@DanBerrios Can you propose a better definition for the term? Thanks

DanBerrios:
A differential expression analysis data transformation is a data transformation that has an objective of achieving a differential expression analysis (determining which biological entities or set of entities are
  likely to have different expression of which genes across one or more conditions).  Typically this involves combining or summarizing gene expression values within entity subgroups and comparing them using statistics.

DanBerrios:
I think this can be closed, no?

PR #967, opened 2018-09-26
Fixes Issues 941 and 942

DanBerrios:


jamesaoverton:
This PR changes the permissions bits on all our files, but the intended changes are minimal.

@DanBerrios: It would be great if you can revise the PR to include only the intended changes. Otherwise @rctauber or I will fix it when we have time.

DanBerrios:
@jamesaoverton @rctauber Why are there a bunch of files listed in the PR with "No changes" below them?  Is it detecting EOL differences or something?  Not sure which files you want me to revise in the PR....  My changes should all be limited to those on the two issues listed.

Public-Health-Bioinformatics:
The permissions were changed on those files, I think that's enough for github to call it a diff.

0  src/ontology/external-byhand.owl 100755 → 100644

DanBerrios:
@jamesaoverton Is there a way to ignore the perm changes? I didn't see how...  Are there other changes you want me to prune out? Which?

jamesaoverton:
As Damion said, the problem is that the file permissions were changed. I don't know why that happened. I haven't seen anything quite like that before.

@rctauber or I will clean it up and make a new PR with just the relevant changes that you made.

jamesaoverton:
Thanks @rctauber. The two new terms 'demultiplexed_sequence_data' and 'multiplexed_sequence_data' would be better in the `sequence-analysis.tsv` template.

rctauber:
I moved those two new terms to the template and assigned the following IDs:
* OBI:0002599 'demultiplexed sequence data'
* OBI:0002600 'multiplexed sequence data'

As I was moving these, I saw that the following axioms had been added to 'material processing':
```
has_specified_input some 'multiplexed sequencing library'
has_specified_output some 'multiplexed sequence data'
```

@DanBerrios did you intend for these to be associated with 'multiplexed nucleotide library sequencing' (OBI:0001959)? Please let me know! Thanks!

jamesaoverton:
@DanBerrios: We'll move ahead with this when we know your answer to @rctauber's question above.

DanBerrios:
@rctauber @jamesaoverton This is what I intended: "Refining existing concept "multiplex nucleotide library sequencing", so that it's output is multiplexed sequence data."  So, no I don't think those axioms belong attached to material processing...they should be attached only to multiplexed nucleotide library sequencing.

rctauber:
@DanBerrios can you confirm that this resolves #941 and #942 now that it's merged? Thanks!

PR #974, opened 2018-10-12
need a place for images belonging to wiki articles

turbomam:
I'm glad to move this to another location if someone want to suggest one


ddooley:
Maybe docs/images/ ?

Issue #1034, opened 2019-06-11
Typo in definition of "achieved by"?


Mttbnchtt:
The definition of "achieved by" is "This relation obtains between a a objective specification and a planned process when the criteria specified in the objective specification are met at the end of the planned process." There is an extra "a" before the first "objective".

Issue #1032, opened 2019-06-11
Typo in "training objective"?


Mttbnchtt:
The 'example of usage' of 'training objective' is "A training objective is fulfilled by e.f. a bioconductor tutorial which instructs the user in the use of a package". Should it be "A training objective is fulfilled by e.g. a bioconductor tutorial which instructs the user in the use of a package" ("e.g." instead of "e.f.")?

bpeters42:
Thanks for catching these - we should fix immediately.

PR #1039, opened 2019-07-08
Fix some typos

rctauber:
See #1032 and #1034


rctauber:
@cstoeckert these typos should be fixed. Could you please review when you get a chance and then I will merge this in? Thank you!

Issue #1050, opened 2019-08-19
unsatisfiable class OBI:comment on investigation


kaiiam:
Greetings OBI team, while building an application ontology which imports form a variety of OBO ontologies, I've run across the following [issue (see it in PMO here)](https://github.com/hurwitzlab/planet-microbe-ontology/issues/1) with the class OBI[:comment on investigation](http://purl.obolibrary.org/obo/OBI_0001898) which when reasoned upon with ELK 0.4.3 becomes inferred as owl:Nothing. 

![image](https://user-images.githubusercontent.com/12255688/63249541-fb996700-c269-11e9-8b14-92c29825dde8.png)

@cmungall pointed out that this issue stems from the combination of a continuant and a process with an AND in the axiom.

DanBerrios:
Discussed on OBI Dev call 08/19/2019.  Assigned to Jie

bpeters42:
The logical definition should probably be changed to: 
'textual entity' and ('is about' some (investigation or ('part of' some investigation' or 'participates in' some 'investigation'))

Will comment on the disjoints separately.

zhengj2007:
The term was added for NIAID BRC standards with description: "Any additional information about the initial specimen collection event, the sample preparation process, or any other aspect of the project worth noting." 

The logical definition proposed by @bpeters42 looks good and fits what NIAID BRC needs. Will update the logical axioms in the OBI.

cmungall:
curious as to how this passed robot checks.

jamesaoverton:
@cmungall Not ROBOT's fault. We haven't been asserting BFO disjoints when building OBI. PMO is getting "continuant DisjointWith 'part of' some occurrent" from somewhere else. I made a separate OBI issue for this #1051.

PR #1053, opened 2019-08-22
updated logical axiom of term OBI_0001898, issue #1050

zhengj2007:
…obi-ontology/obi/issues/1050


jamesaoverton:
Thanks @zhengj2007. I check this locally by adding the axiom

    continuant DisjointWith 'part of' some occurrent

then ran the reasoner, and found another problem with OBI:0002053:

    'assay measuring qualitiative binding of a B cell epitope:antibody complex'
 and ('has part' some microarray)

I think it should be 'has part' some 'microarray assay'. I'll update this PR to fix that problem too.

Issue #1021, opened 2019-05-14
NTR: medication timepoint


johnwjudkins:
NTR: medication timepoint

Definition: a time measurement datum that specifies the point in time that a human participant under investigation self-administers a medication material

Parent: time measurement datum

mbrochhausen:
I might be wrong, but I think you are not really looking at a time measurement datum here (definition: A scalar measurement datum that is the result of measuring a temporal interval.) What you seem to be looking for seems to be more the identifier of a specific time point (think Star Trek star dates). These aren't measurements. The are names we stick on specific periods of time (notable the name of one time slice can be different in the Central time zone and in the Eastern time zone). Just food for thought.

ddooley:
A number of OBI folks are leaning towards adding (see https://github.com/obi-ontology/obi/issues/981)

    instantaneous time value specification
        'has specified value' some (xsd:dateTime or xsd:dateTimeStamp)"

So then above defn. could be: 

    an instantaneous time value specification that specifies the point in time that a human participant under investigation self-administers a medication material

johnwjudkins:
Thank you both for your suggestions. I agree the term is an identifier/specifier of a time point rather than a measurement. @ddooley , that definition works for me (even if we just call it a value spec w/o the phrase "instantaneous time").

ddooley:
About whether it is a measurement or not - consider that all such recorded times (not predictions or simulations though) are achieved by noting the digits provided by an analog or digital clock device. Whatever it spits out is a measurement, regardless of the underlying scheme the numbers present, in the same way as measuring "20g" mass. Experimentally one may later discover an error occurred with the clock apparatus just like any other device. It suffers the vagaries of measurement.

ddooley:
I should add that I think the primary semantic defn. for "medication timepoint" should be a value specification as above.  Its just that one can then additionally layer on whether it is a measurement, recommendation, estimate etc. if needed.

jamesaoverton:
Discussed on the OBI call 2019-06-17.

This should go in the [`value-specification.tsv`](https://github.com/obi-ontology/obi/blob/master/src/ontology/templates/value-specifications.tsv).

I think that this term is fine, but there are still decisions to make about #981.

beckyjackson:
This is a term for a record of something that is *already completed*, as in the medication has already been administered.

Action item: @johnwjudkins to create PR

Issue #1020, opened 2019-05-14
NTR: medication dose


johnwjudkins:
NTR: medication dose

definition: a dose that measures the quantity of a medication material administered to an organism

parent: dose (OBI_0000984)

jamesaoverton:
Discussed on OBI call 2019-06-10.

This makes sense as a child of 'dose', but like #1021 it would be more useful to have as a value specification going forward.

'medication material' and 'medication role' are both in OBIB

johnwjudkins:
My proposed definition for this term as a value specification:

label: medication dose value specification
definition: a scalar value specification that specifies the value of the quantity of a material entity administered to an organism as a treatment

(alternatively, a scalar value specification that specifies the value of the quantity of a medication material administered to an organism)

BTW this and 'medication timepoint' are intended to be about already-administered treatments rather than prescribed treatments.

beckyjackson:
It may make sense to subclass this to specify the measurements (g, ml, etc...)

Action item: @johnwjudkins to create PR for this

PR #1061, opened 2019-10-14
Jwj20191008

johnwjudkins:
NTRs: #1037, #1021, #1020.

Slight grammatical change to case-control study design definition: "What distinguishes the two types of roles is an 'outcome' which is associated with participant that have the case role, while a participant having the control role do not." is changed to "What distinguishes the two types of roles is an 'outcome', which is associated with participants that have the case role _but not associated with participants that have the control role_."


beckyjackson:
`OBI:0002619` ID will conflict with PR #1060 as it uses 0002619 through 0002623 - can I update the ID for 'case control study' in this PR to be `OBI:0002624`? If you're already using the current ID, I will update #1060.

beckyjackson:
Thanks!

@jamesaoverton - I updated the ID and generated the modules. Should be good to go.

Issue #1067, opened 2019-10-30
Issue with 'trial statistician role' axiom


beckyjackson:
'trial statistician role' (`OBI:0000234`) has the axiom:
```
'has part' some 
    (realizes some 
        (concretizes some 'clinical study design')) 
 and ('realized in' only investigation)
```

'realizes' has a domain of 'process' which means that 'trial statistician role' has part some process. Since this is a role (a continuant), it cannot have a process (occurrent) part.

beckyjackson:
I think you could say:
```
'realized in' only 
  (investigation and
    ('has part' some 
      (realizes some 
        (concretizes some 'clinical study design'))))
```

To say that the 'trial statistician role' is realized in an investigation that has a part where a 'clinical study design' is executed. I could be incorrect, though.

bpeters42:
This looks good to me

DanBerrios:
Discussed on call 11/4/19.

beckyjackson:
This is fixed in the new release.

PR #1068, opened 2019-10-30
[WIP] Create an integration test

beckyjackson:
1. Create an `obi_base.owl` product (see https://github.com/ontodev/robot/pull/570) \*
2. Retrieve BFO classes, RO core, and IAO metadata
3. Merge these together and run HermiT reasoner
4. If reasoning fails, dump unsatisfiable module (`build/unsatisfiable.owl`) & create explanations (`robot explain`) for unsatisfiable classes (`build/explain/*.md`)

\* This may not be complete as it is only including stuff in the OBI namespace. If there are other classes that should be included in the "base" product (e.g., stuff OBI has adopted), we can add those in.

This is not run on Travis due to use of the HermiT reasoner. I have included it as part of the release (`make all`) and it can be run on its own using `make verify-integration`.

Right now it fails with one unsatisfiable class (see #1067), but you can see a full explanation in `build/explain/OBI_0000234.md` (after running this test). This might be a little overkill, but it's a cool use of `robot explain` 🙂


jamesaoverton:
@beckyjackson On Monday's call we agreed to spend some time improving OBI's ROBOT report and OBO Dashboard score. For that we need a proper `obi-base.owl` file. I'd appreciate your help with that, either by reviving this PR or creating a new one.

PR #1074, opened 2019-11-04
Berrios

DanBerrios:
Resolves #1069

PR #1075, opened 2019-11-05
Add a device and an assay, resolves #1069

### LINKED ISSUE #1069: NTR: Illumina NovaSeq 6000 Sequencing Platform
Project: ENCODE 

Illumina's high throughput sequencing instrument, the NovaSeq 6000 was released in 2017 and is most likely the successor of Illumina's HiSeq models. The NovaSeq leverages cost effectiveness with their highest output yet (up to 6 Tb of data). 

NTR: "Illumina NovaSeq 6000"

- Alternative terms: NovaSeq 6000
- Textual definition: A DNA sequencer which is manufactured by the Illumina corporation, with two flow cells and an output of up to 6000 Gb (32-40 B reads per run). The sequencer utilizes synthesis technology and patterned flow cells to optimize throughput and even spacing of sequencing clusters. 
- Definition source for the textual definition: http://www.illumina.com/systems/sequencing-platforms/novaseq.html
- Logical definition (or parent term):
- Example of usage: Lei, Y., Zhang, J., Cao, L., Zhao, R., Feng, J., & Li, B. (2018). Draft genome sequence of _Xenophilus_ sp. E41 isolated from an activated sludge reactor treating wastewater with high cephalexin concentration. Journal of Global Antimicrobial Resistance (2018), PMID: 30316965
- Term editor: Bonita Lam

jamesaoverton:


jamesaoverton:
Copied from #1074 which has merge conflicts.

There's another problem here: a new assay is defined in `obi-edit.owl`, but that should be a violation of the `assay-violation-edit.owl` rule. @beckyjackson Please look into that.

jamesaoverton:
Once we figure out why the rule is not failing, the new assay should be moved into `assays.tsv`.

beckyjackson:
That new assay isn't a direct child of assay, and even though the query uses `rdfs:subClassOf*` the axiom linking the parent class up to assay is located in the assays module. So this rule would only catch assays that are directly asserted to be assays in `obi-edit.owl`.

I'm not sure the best way to handle this... a simple fix would be looking for any label that ends with the word "assay". A more complex fix would be to remove all annotations from the assays module, merge that in, and then look for any descendants of assays that still have annotations. Any other ideas?

As a side note, I'm seeing this in Protege but I don't see it on the PR:
Screen Shot 2019-11-05 at 7 30 07 AM

jamesaoverton:
Ok, the assay term is from the previous PR #1071 and should not be in this PR. But for #1071 I should have put it in `assays.tsv` and the check didn't catch that.

beckyjackson:
To confirm, this PR *adds* the term 'Illumina NovaSeq 6000' (`OBI:0002630`) and *moves* 'direct RNA-sequencing assay' (`OBI:0002629`) to the assays template. 'direct RNA-sequencing assay' was duplicated as `OBI:0002631` so I have deleted `OBI:0002631`.

Was there supposed to be a second device?

Also, I ran `make fix` which is why there are some other changes showing up in the assays template. Weird EOL things.

jamesaoverton:
This is right. Thanks!

PR #1078, opened 2019-11-07
Move all study designs to template

beckyjackson:
Builds on #1077 and adds a test to make sure that no study design is asserted in `obi-edit.owl`.

The assays module was updated because a study design snuck in there, so it was moved to the study design template.

There are some potential problems to review for the study design module:

Assay for 'organism status comparison design'
---

```
('is concretized as' some 
    ('realized in' some 
        (assay and 
            (has_specified_input some 
                ('pre-mortem specimen' or 'post mortem specimen')))))
```

This does not match the assay axiom pattern: `'is concretized as' some ('realized in' some ('has part' some %))`. To match this pattern, it would look like:
```
('is concretized as' some 
    ('realized in' some
        ('has part' some 
            (assay and 
                (has_specified_input some 
                    ('pre-mortem specimen' or 'post mortem specimen'))))))
```

Do we need to review this pattern, or is this axiom incorrect?

Equivalent statement for 'protein binding site identification design'
---

'protein binding site identification design' is equivalent to `('has part' some 'protein and DNA interaction identification objective')`. I think this should either be moved to a subclass, or changed to:
```
'study design' and 
    ('has part' some 'protein and DNA interaction identification objective')
```

Definition for 'hyper graeco latin square design'
---

The current definition is just:
> PRS to do


jamesaoverton:
I's like to move ahead with this PR. Is there anything blocking it?

src/ontology/templates/assays.tsv
```diff
@@ -307,6 +307,8 @@ OBI:0002441	physical examination of individual			"An assay that produces a descr
...
+OBI:0002447	suppression subtractive hybridization		SSH|subtractive hybridization	"An artificially induced nucleic acid hybridization that is performed to compare gene expression in different cell or tissue types based on normalization and suppression, which creates a subtracted cDNA or genomic DNA library."	PMID:14970460			Rebecca Jackson			subclass		artificially induced nucleic acid hybridization												
+OBI:0002448	differential screening hybridization		DSH|differential screening	An artificially induced nucleic acid hybridization that is performed to identify differentially expressed genes through hybridization of cDNA probes.	PMID:25472628			Rebecca Jackson			subclass		artificially induced nucleic acid hybridization
```
jamesaoverton: These two are not assays? I see them under 'material processing'. There should be a check to catch this.

PR #1087, opened 2019-11-25
Circulome-seq as a synonym for OBI:0002144, #872

turbomam:


jamesaoverton:
See #872. This would be better in the `assays.tsv` file, not in `obi-edit.owl`.

turbomam:
this could be deleted in favor of #1104

Issue #863, opened 2017-09-11
polyA+ & polyA- RNA-seq distinctions


jahilton:
Hello,
The ENCODE project is exploring how to distinguish various RNA-seq assays, including those that involve steps in the protocol to enrich, or deplete, polyAdenylation. Would these be appropriate terms?

editor preferred term: polyA-selected RNA sequencing assay
alternate terms: polyA RNA-seq, polyA mRNA RNA-seq, mRNA-seq
textual definition: An RNA-seq assay to study transcriptomes through the enrichment of polyadenylated transcripts or removal of ribosomal RNA prior to high-throughput sequencing.
definition source for the textual definition: https://bmcgenomics.biomedcentral.com/articles/10.1186/1471-2164-15-419
logical definition: RNA-seq assay --> polyA RNA-seq

editor preferred term: polyA-depleted RNA sequencing assay
alternate terms: polyA-depleted RNA-seq
textual definition: An RNA-seq assay to study non-polyadenylated transcripts, such as, but not exclusive to unspliced and circular isoforms, through the depletion of polyadenylated transcripts prior to high-throughput sequencing.
definition source for the textual definition: http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0030733, http://www.tandfonline.com/doi/full/10.1080/15476286.2015.1110676
logical definition: RNA-seq assay --> polyA-depleted RNA-seq

Thanks!

turbomam:
Hi, Jason.  This is a good request.  I'd like to discuss it during the OBI phone call at noon EDT on this Monday, April 30th.  I assume the group will approve it, but I'll be suggesting that we follow some OBI templates and define these assays in terms of what their specified input is.  That may require creating a few additional terms (which may or may not interest you.)

Would you be willing to join the call to make sure your needs and perspective are understood?

see http://obi-ontology.org/#contact-us

jahilton:
Yes, I think the ENCODE DCC will have some representation on next week's call.
Thank you.

turbomam:
I just noticed that you have several additional term requests, all from mid to late 2017.  I'll try to orient myself regard the others, in case you or a colleague want to go over them, too.

turbomam:
OBI has related terms using at least two notations: " polyA..." and "poly(A)..."

Should we make these consistent?

turbomam:
[...]

PR #1088, opened 2019-11-25
polyA+ & polyA- RNA-seq distinctions #863

turbomam:
I edited `assays.tsv` in Excel and `obi-edit.owl` in Protege, but `git status` says I modified

- src/ontology/modules/assays.owl
- src/ontology/obi-edit.owl
- src/ontology/templates/assays.tsv


jamesaoverton:
If a term belongs in a template, then all its annotations should be in that template as well. So there should be no need to use Protege or touch `obi-edit.owl` in this case.

Unfortunately, a bunch of special characters in `assays.tsv` have been replaced with `???`.

Is the desired change just to add an alternative term?

turbomam:
closing in favor of #1105

PR #1118, opened 2020-03-08
2 new terms and a definition touchup

ddooley:
planned process: "processual entity" -> process
NTR: enrichment culture medium
NTR: case series design


jamesaoverton:
Please move the 'case series design' to the [study-design.tsv](https://github.com/obi-ontology/obi/blob/master/src/ontology/templates/study-designs.tsv) template.

ddooley:
So it turns out "planned process" gets two definitions, one defined in OBI, and the other coming in from IAO.  If they aren't the same string of text, which they aren't anymore, then this throws a multiple definitions error.  It would seem the IAO definition must be corrected too, or more elegantly, blocked from import?

jamesaoverton:
Closing in favour of #1131

PR #1119, opened 2020-03-09
NTR request "massively parallel reporter assay"

ddooley:
OBI:0002675	massively parallel reporter assay
New term is last one in sheet.
One issue is that my edit using MS Excel seemed to have a very inneficient diff - all rows of previous file removed, and all new ones added + new row.  Possibly a PC/mac line end save issue?


jamesaoverton:
@ddooley Yes, line endings are a pain. The README has information: https://github.com/obi-ontology/obi#line-endings

Can you please try that and see if it fixes the diff?

PR #1144, opened 2020-03-31
Update term import

### LINKED ISSUE #1135: Review REO terms for adoption into OBI
OBI has used some Reagent Ontology terms, but REO was never actually registered. We should review the list of terms and consider adopting them into OBI.

Repository: https://github.com/tis-lab/reagent-ontology

OBO-level issue: https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1042
Previous OBI issue: #1015

@mbrush says:

> The best file to explore if you want to poke around the ontology itself is the pre-reasoned version here. The central hierarchy rooted at the 'reagent' class is what you will want to explore - but it is built only upon running a reasoned over the ontology to collect all classes with a reagent role in one place. This pre-reasoned version is useful because ReO's baroque axiomatization of classes in the ontology make reasoning in Protégé prohibitively slow (at least with the most common Protege reasoner plugins).

We should try to create a `reagents.tsv` ROBOT template. @beckyjackson Please try running your reverse-template code on REO, when you get a chance.

### LINKED ISSUE #1141: Deprecated GO term
OBI imported GO term, [neuron part](http://purl.obolibrary.org/obo/GO_0097458), which is used in defining [single-unit recording](http://purl.obolibrary.org/obo/OBI_0002187).

Current [neuron part](http://purl.obolibrary.org/obo/GO_0097458) is deprecated in GO and suggest to consider using CL term [neuron](http://purl.obolibrary.org/obo/CL_0000540). 

Shall we make corresponding updates in OBI?

zhengj2007:
Solve issues related to term imported from external resources
resolves #1141 deprecated GO term
resolves #1135 not registered REO terms


jamesaoverton:
Thanks @zhengj2007. I noticed that REO:0000280 was still in `obi-edit.owl`, so I replaced it with OBI:0002744. I also added REO:0000171 and REO:0000280 to `obsolete.tsv` with their replacements noted, since OBI users may have used those terms.

Can you please double-check this for me?

zhengj2007:
> Thanks @zhengj2007. I noticed that REO:0000280 was still in `obi-edit.owl`, so I replaced it with OBI:0002744. I also added REO:0000171 and REO:0000280 to `obsolete.tsv` with their replacements noted, since OBI users may have used those terms.
> 
> Can you please double-check this for me?

Looks good to me. Thanks!

PR #1146, opened 2020-04-03
Add script to sort templates and standardize quoting and line endings

jamesaoverton:
Git works best when each commit changes only the things that really matter. Then we can `git diff` to see what really changed, and look through the history of every file. That has mostly been working for our OWL files for the past few years.

Unfortunately, we keep seeing spurious changes in our template files. Different programs save TSV differently, with different quoting rules and different line endings. People tend to add their terms to the end of the template, but it's nice to keep the templates sorted by ID.

This PR adds a Python script to read all the templates, sort them by ID, and save them again with consistent quoting and line endings. I've applied this script to all our templates, so this commit is a big mess that will be an "event horizon" in our git history. But fixing this now means we will have clean diffs and history going forward.

Note that these changes make no difference to the OWL files generated from the templates. The only change I made to the content was removing some weird quoting for a single example of usage in the `value-specifications.tsv`.


jamesaoverton:
Discussed on the OBI developer call 2020-04-06. We will go ahead with this sorting by ID. We won't attempt the tree sort alternative that I presented. I can provide that script when it's ready so people can try tree sort locally.

PR #1160, opened 2020-04-24
Add a check for bad IRIs

### LINKED ISSUE #1158: Add check for bad IRIs
#1156 slipped past all our tests. We need a new check to catch bad IRIs.

beckyjackson:
Resolves #1158 

Check that all referenced IRIs have a valid namespace.


jamesaoverton:
Thanks! There might be some other cases we want to check. Let me think about this a little longer.

PR #1161, opened 2020-04-27
Add all obsolete classes to template

### LINKED ISSUE #1159: Move all obsolete terms to `obsolete.tsv`, add check
I was surprised to find lots of obsolete terms in `obi-edit.owl`. They should all be in `obsolete.tsv`, and we should have a check to make sure that none remain in `obi-edit.owl`.

beckyjackson:
Resolves #1159

Row 59 of the `obsolete.tsv` template has a multi-line annotation which might cause the GitHub rendering of TSVs to break, but the template loads properly... we could always combine this into one line if that's a problem.


jamesaoverton:
I don't want newline in the TSVs. I'd prefer to escape them with "\n" automatically in the new `sort-templates.py`. What does ROBOT do with "\n" in templates?

jamesaoverton:
I'd also like a check to ensure that `obi-edit.owl` does not contain any obsolete terms.

Most of our templates form a branch of the tree, and each of those should be checked. Maybe  generalize `assay-violation-modules.rq`?

beckyjackson:
Travis is failing with:
```
FAIL Rule src/sparql/missing-logical-definition-violation.rq: 7 violation(s)
s,label,type
http://purl.obolibrary.org/obo/OBI_0000624,Bruker SampleJet system,http://www.w3.org/2002/07/owl#Class
http://purl.obolibrary.org/obo/OBI_0000516,NMR probe,http://www.w3.org/2002/07/owl#Class
http://purl.obolibrary.org/obo/OBI_0000555,autosampler,http://www.w3.org/2002/07/owl#Class
http://purl.obolibrary.org/obo/OBI_0000531,Bruker CryoProbe,http://www.w3.org/2002/07/owl#Class
http://purl.obolibrary.org/obo/OBI_0000612,Bruker MATCH tube holder system,http://www.w3.org/2002/07/owl#Class
http://purl.obolibrary.org/obo/OBI_0000590,Bruker 1mm MicroProbe,http://www.w3.org/2002/07/owl#Class
http://purl.obolibrary.org/obo/OBI_0000517,NMR magnet,http://www.w3.org/2002/07/owl#Class
```

But I don't get this locally. Any ideas on why this is happening? I believe these are devices, which may point to something...

jamesaoverton:
It's passing locally for me too. Not sure what the problem is.

jamesaoverton:
Now I'm pretty confident that it's a difference between `sed` implementations on macOS and Linux. Please use Python to do this instead.

beckyjackson:
Hmm... it still doesn't seem to like it.

jamesaoverton:
I just don't get it. Let's remove the newlines from `obsolete.tsv`, delete the script from this PR, and worry about this problem the next time we see it.

PR #1162, opened 2020-05-04
Create biopsy.tsv

ddooley:
A list of 15 biopsy terms, a number of which exist to some extent in other ontologies:
**biopsy**	Replaces http://purl.obolibrary.org/obo/ERO_0001334
**image guided biopsy**	"http://purl.obolibrary.org/obo/OMIT_0027992
and http://purl.obolibrary.org/obo/NCIT_C15652"
**computed tomography assisted biopsy**	http://purl.obolibrary.org/obo/NCIT_C137909
**fine needle aspiration biopsy**	http://purl.obolibrary.org/obo/OBIB_0000210
**core needle biopsy**	http://purl.obolibrary.org/obo/OBIB_0000211
**abdominal wall fat pad biopsy**	http://purl.obolibrary.org/obo/ERO_0001335 "fat biopsy"
**tail biopsy**	http://purl.obolibrary.org/obo/ERO_0001339
**bone marrow aspiration biopsy**	http://purl.obolibrary.org/obo/ERO_0001761
**skinpunch biopsy**	http://purl.obolibrary.org/obo/ERO_0001336


jamesaoverton:
Sorry, I can't read this table. Please:

- revise the header to use simple column names: no examples, no newlines
- revise the column order to more closely match other templates: ID, label, other annotations, logic
- if possible, please avoid newlines inside cells
- please use single quotes when referring to a term in an annotation; double quotes are being escaped with double- and triple- double quotes
- there seem to be rows without IDs: please remove these
- once that's done, please run `make sort`

ddooley:
Oops. Sorry, I totally forgot about cleaning up curator comments.  New table coming shortlly.

jamesaoverton:
Thanks, that's looking much better. I'd still like a few changes so that this works like other OBI templates.

1. I'd prefer this column order: Ontology ID | label | has curation status | alternative term | definition | definition source | example of usage | term editor | ontology term requester | class type | parent class | material anatomical entity | input device | output entity

2. If the "curation notes" column doesn't go in the OWL (no ROBOT template string), then I think that it should not be in the template

3. Please use labels instead of CURIEs throughout. You'll have to add a few UBERON terms to `src/ontologies/OntoFox_inputs/Uberon_input.txt`

ddooley:
Ok.  New sheet references one other OBI term for which a NTR is just opened - OBI:0002679 for "vacuum-assisted biopsy needle", ok?

ddooley:
ok, done!

jamesaoverton:
Ok, just one more problem: The ROBOT template command is failing because OBI does not contain a term 'multicellular organism'. But since the column is called "material anatomical entity", don't you want to just use http://purl.obolibrary.org/obo/UBERON_0000465 'material anatomical entity' instead?

ddooley:
You are so right - since that column should point to an anatomical part or region. I'll adjust.

jamesaoverton:
I think this is good to go. Do you agree @ddooley?

ddooley:
Yes, thanks for your help, James!

PR #1165, opened 2020-05-11
Update obi-edit.owl to fix #1120 (supplies --> provisions)

### LINKED ISSUE #1120: Change label of OBI object property: supplies
Discussed on OBI March 9, 2020 call in context of ROBOT report.
OBI relation 'supplies' http://purl.obolibrary.org/obo/OBI_0000646 has the same label as RO 'supplies' http://purl.obolibrary.org/obo/RO_0002178
OBI's term is about supplying material entities (as a service) whereas RO's term is about anatomy and is a child of connected. We propose changing OBI's label to reflect this distinction.

bpeters42:
Changed label, definition, alternative label and editor note to fix #1120.


jamesaoverton:
This is a good Pull Request. Thanks!

It would be a little better in the future if you used the latest version of Protege, since there are some tiny differences in the output with the version you're using from 2018.

Issue #864, opened 2017-09-12
NTRs: ATAC-seq assays


jahilton:
editor preferred term: single-nuclei ATAC-seq
alternate terms: snATAC-seq
textual definition: Analysis of the transposase-accessible-chromatin in single nuclei isolated from frozen tissue samples through a protocol designed to maximally preserve nuclei integrity during sample processing and optimize transposase-mediated fragmentation of chromatin in individual nuclei. snATAC-seq can be used to dissect heterogeneity and delineate gene regulatory sequences in complex tissues, and uncover both the cell types and the regulatory elements in each cell type without prior knowledge.
definition source for the textual definition: http://www.biorxiv.org/content/early/2017/07/06/159137
logical definition: ATAC-seq --> snATAC-seq

editor preferred term: single-cell ATAC-seq
alternate terms: scATAC-seq
textual definition: A robust method for mapping the accessible genome of individual cells by assay for transposase-accessible chromatin using sequencing (ATAC-seq) integrated into a programmable microfluidics platform in order to provide insights into cell-to-cell variation in accessibility profiles.
definition source for the textual definition: http://www.nature.com/nature/journal/v523/n7561/full/nature14590.html
logical definition: ATAC-seq --> scATAC-seq

gabdank:
I was wondering if there was made in progress with this new term request?
We would like to update the term request for the snATAC-seq

editor preferred term: single-nucleus ATAC-seq
alternate terms: snATAC-seq
textual definition: A combinatorial barcoding-assisted single-cell assay for transposase-accessible chromatin that is optimized for use on flash-frozen primary tissue samples. snATAC-seq can be used to dissect cellular heterogeneity and delineate cell-type-specific gene regulatory sequences in snap-frozen primary tissues.
definition source for the textual definition: https://www.nature.com/articles/s41593-018-0079-3
logical definition: ATAC-seq --> snATAC-seq

mkuhring:
Pushing #864 and #995. Single cell technology should really be part of OBI, probably by yesterday.
What is keeping these terms from being implemented for two years now? These term submissions look complete to me. Definition sources can be also used as examples.

turbomam:
Some of these requests have admittedly slipped through the cracks. I can do a few this week, but will need to prioritize relative to request from @carrieadavis

bonitalam:
Hi @turbomam, on behalf of ENCODE - I believe the ATAC-seq (ATAC-seq and snATAC-seq) requests have higher priority.

[...]

PR #1188, opened 2020-07-18
ordinal value specification and singleX ATAC

### LINKED ISSUE #862: ordinal value specification
Due to the interest in ordinal value specifications, I'd like to take a small but concrete step as soon as possible.  Let's discuss this on Monday September 11th.

create "ordinal value specification" as a subclass of "categorical value specification", <http://purl.obolibrary.org/obo/OBI_0001930> with URI <http://purl.obolibrary.org/obo/OBI_0002930>

proposed definition:  "a categorical value specification in which the available categories have a natural order and the distances between the categories is not known" from https://en.wikipedia.org/wiki/Ordinal_data

example of use:  "A RNA integrity number of 1 indicates RNA that is more highly degraded than RIN 9"

Future steps:  
 - what kind of class would contain multiple ordinal (or categorical) value specifications like 'infant', 'neonate' and 'toddler'?  I.e., how to express which ordinal values are legal in a given context?
 - what classes or properties to create, in order to express the rank of a ordinal value specification ('infant' = 2, 'neonate' =1, 'toddler'=3)?

turbomam:
Fixes #862 and #864


jamesaoverton:
Travis is failing because the new assays should be entered into the `assay.tsv` and not `obi-edit.owl`.

turbomam:
I think [ordinal value specification](http://purl.obolibrary.org/obo/OBI_0003001) for #862 got in here, too.

jamesaoverton:
The diff for the `assays.tsv` file is a mess. Can you please run the [sort-templates.py](https://github.com/obi-ontology/obi/blob/master/src/scripts/sort-templates.py) script to clean it up?

jamesaoverton:
I fixed merge conflicts and assigned IDs using the [OBI Term ID Reservation](https://docs.google.com/spreadsheets/d/1tpDrSiO1DlEqkvZjrDSJrMm7OvH9GletljaR-SDeMTI/edit#gid=224833299) sheet:

OBI:0002756 ordinal value specification
OBI:0002757 single-nucleus ATAC-seq
OBI:0002758 single-cell ATAC-seq

Issue #1189, opened 2020-07-20
Term request: estimated


reality:
Some discussion here https://github.com/pato-ontology/pato/pull/323 . Will open a pull request shortly!

jamesaoverton:
Thanks for the term request and PR. I agree that OBI is the right place to model estimates and approximations.

In OBI/IAO we have created a number of different 'data item' classes, which are all the outputs of different measurement processes (i.e. assays). However we found that we wanted to reuse these classes for more than just measurements: predictions, simulations, settings, etc. So we changed our modelling to focus on 'value specifications', which are neutral with respect to the process that generated them. Here's a good overview: http://obi-ontology.org/docs/data-intro/. We have more detailed, developer-focused documentation if you prefer.

Does this value specification approach fit your use case for estimations and approximations?

jamesaoverton:
I think I addressed the question, but if not, please reopen.

PR #1190, opened 2020-07-20
add estimated value class OBI:0002749 for #1189

reality:
#1189 

Have also added an entry to the spreadsheet!


jamesaoverton:
I'll mark this as a draft until I hear back about my comment on #1189.

Issue #1202, opened 2020-08-10
NTR: "presence or over-threshold value specification"


turbomam:
As discussed at noon Eastern time today

label: "presence or over-threshold value specification"
subClassOf: "ordinal value specification"
definition: "An ordinal value specification that indicates that either a) an analyte or mesaurand is present in an evaluant, or b) that the amount of analyte/measurand detected is over some predetermined threshold."
Definition source: LOINC User's Guide, Version 2.68, PUBLISHED JUNE 2020, page 32

PR #1207, opened 2020-08-11
LOINC imports

turbomam:


turbomam:
@jamesaoverton I apologize for pestering. I was able to find some errors I made in my submission, but I am still getting the following and don't understand. I ran `make all` on my laptop and didn't see any complaints.

```Bash
java -jar build/robot.jar --prefix "REO: http://purl.obolibrary.org/obo/REO_" merge \
--input src/ontology/obi-edit.owl \
template \
--template src/ontology/templates/sequence-analysis.tsv \
annotate \
--ontology-iri "http://purl.obolibrary.org/obo/obi/dev/modules/sequence-analysis.owl" \
--output src/ontology/modules/sequence-analysis.owl
Could not load imported ontology: <http://purl.obolibrary.org/obo/ro/annotations.owl> Cause: connect timed out
```

src/ontology/OntoFox_inputs/OBO_input.txt
```diff
@@ -157,6 +157,44 @@ http://purl.obolibrary.org/obo/PR_000025475
 http://purl.obolibrary.org/obo/PR_000025477
 http://purl.obolibrary.org/obo/PR_000025478
 http://purl.obolibrary.org/obo/PR_000003252
+http://purl.obolibrary.org/obo/DRON_00018971 # adalimumab for TURBO assays
```
jamesaoverton, line 160: This seems to be out of place.

src/ontology/OntoFox_inputs/NCBITaxon_input.txt
```diff
@@ -34,6 +34,18 @@ http://purl.obolibrary.org/obo/NCBITaxon_11103
 http://purl.obolibrary.org/obo/NCBITaxon_10376
 http://purl.obolibrary.org/obo/NCBITaxon_10358
 http://purl.obolibrary.org/obo/NCBITaxon_160
+http://purl.obolibrary.org/obo/NCBITaxon_1 # root for TURBO assays
```
jamesaoverton, line 37: Can you use OBI 'organism' instead? That's what we've done elsewhere.

src/ontology/OntoFox_inputs/UO_input.txt
```diff
@@ -16,6 +16,9 @@ http://purl.obolibrary.org/obo/UO_0000105
 http://purl.obolibrary.org/obo/UO_0000109
 http://purl.obolibrary.org/obo/UO_0000270
 http://purl.obolibrary.org/obo/UO_0000280
+http://purl.obolibrary.org/obo/UO_0000061 # unit of molarity for TURBO assays
```
jamesaoverton, line 19: These three should be in just one of these UO files. It just depends on whether we want it in OBI as a class or as an instance.

jamesaoverton:
I made some specific comments. One general comment: the various "release" files should not be included in the PR. This means `obi.owl` and the files in `views/`. I just update these when we make a new release. You only need to commit stuff under `src/`. This is probably not clear in the REAMDE...

jamesaoverton:
This is a (temporary) network error "Could not load imported ontology: <http://purl.obolibrary.org/obo/ro/annotations.owl> Cause: connect timed out". When I see one of these, I click the "Restart build" button.

jamesaoverton:
Is this PR supposed to add one term for #1202, or is it supposed to update hundreds or imports for the LOINC work?

turbomam:
Well, probably both. Is there something I can do to split those two things
up?

jamesaoverton:
I moved the work on #1202 to new PR #1212.

To update imports you need to run `make imports`. Then the updated files in `src/ontology/OntoFox_outputs` should be added to this PR.

However it will not be that simple. Imports go out of date as the upstream ontologies change. When I ran this locally and checked the diffs, it looked like GO had obsoleted some terms OBI was using, for starters. I'm also surprised that you need *so* many ChEBI terms, and I have other questions about the new imports.

Overall, it might be better to update the update the imports, one-by-one in separate PRs, without your newly imported terms, and resolve any problems with the upstream ontologies. Then it would be easier to import your new terms.

I'm sorry that this is difficult. There's about a thousand moving pieces.

jamesaoverton:
Since this is not ready to go, I converted to a draft to prevent an accidental merge.

jamesaoverton:
I updated the import files (without adding any new imported terms) in #1216, and I updated this branch with all the latest changes. So you can continue your LOINC work on this branch.

jamesaoverton:
@turbomam What's the status of this? Should we mark it as Draft again?

Issue #1203, opened 2020-08-10
NTR: Immunoassay process


turbomam:
As discussed at noon Eastern time today. I would like to add this soon, in support of a Covid assay that's common at the University of Pennsylvania Healthcare System, but will I wait a few days for community feedback.

label: "Immunoassay process"
subClassOf: "assay"
definition: "An assay that detects the linking of antigens and antibodies via special signaling mechanisms, such as EIA, ELISA, chemiluminescence, etc. that produces one measure (quantitative or qualitative) of the analyte of interest."
Definition source: LOINC User's Guide, Version 2.68, PUBLISHED JUNE 2020, page  47

turbomam:
LOINC indicates that this class would not include flow cytometry or any technique that can determine the localization of an analyte, such as peroxidase immune staining or indirect immunofluorescence (IFA)

turbomam:
Potential label:

Single-measure binding of antibody with reporter to antigen assay

jamesaoverton:
Discussed on the 2020-08-17 call. Bjoern doesn't love "single-measure" but we could not think of something better. @turbomam will go ahead with this.

cstoeckert:
This is in OBI.

PR #1214, opened 2020-08-17
Single-value binding of antibody with reporter to antigen assay #1203

turbomam:


jamesaoverton:
You should run `git diff` before committing (or pushing) your changes -- that's step 5 in this part of the README:

https://github.com/obi-ontology/obi#making-changes

The `git diff` will show that every line of the `assays.tsv` template has changed. That's bad. `make sort` should fix it. Without a good diff, it's very difficult to review a PR.

turbomam:
Sorry about the garbled assays template, @jamesaoverton . Thanks for the feedback. I just ran `make sort`, but at least in my local environment there's still pages of different lines. I committed locally and can push now if that's best.

There was a message on this page about my branch being out of sync with master, so I impulsively clicked a button with a label like "refresh".

jamesaoverton:
`git diff` compares your unstaged changes to the HEAD (last commit). In this case the last commit has bad line endings, so running `make sort` and `git diff` will show that all the lines have changed (i.e. been fixed).

What we really want is to compare the results of `make sort` to the `master` branch. So the right command is `git diff master -- src/ontology/templates/assays.tsv`.

When I pulled this branch and ran `make sort` and `git diff master -- src/ontology/templates/assays.tsv` the results looked good. So I pushed them. You should now see a clean diff on the "Files changed" tab. Please confirm that it has the changes you want.

turbomam:
Looks good. Thanks. Do I need to wait for anything before I add my LOINC-inspired assays into the OBI assay template?

Issue #1215, opened 2020-08-18
Replace obsolete GO 'core promoter binding'


jamesaoverton:
GO has replaced 'core promoter binding' http://purl.obolibrary.org/obo/GO_0001047 with 'core promoter sequence-specific DNA binding' http://purl.obolibrary.org/obo/GO_0001046

'core promoter binding' was defined as "Interacting selectively and non-covalently with the regulatory region composed of the transcription start site and binding sites for the basal transcription machinery. Binding may occur as a sequence specific interaction or as an interaction observed only once a factor has been recruited to the DNA by other factors."

'core promoter sequence-specific DNA binding' is defined as "Interacting selectively and non-covalently with a sequence of DNA that is part of a core promoter region. The core promoter is composed of the transcription start site and binding sites for the RNA polymerase and the basal transcription machinery. The transcribed region might be described as a gene, cistron, or operon."

According to [Ontobee](http://www.ontobee.org/ontology/OBI?iri=http://purl.obolibrary.org/obo/GO_0001047) it looks like we just use 'core promoter binding' in OBI to define 'promoter activity detection by reporter gene assay' http://purl.obolibrary.org/obo/OBI_0000913

@rvita @bpeters42 Can we replace 'core promoter binding' with 'core promoter sequence-specific DNA binding'?

turbomam:
whoa, I have no idea how that got in there. I'll check to see if I really need it

jamesaoverton:
@turbomam Sorry, I don't understand your comment. I added this import four years ago for use in an assay for IEDB. But now it's obsolete, so we need to update it.

turbomam:
yeah, I should have read the whole post. Sorry about that.

jamesaoverton:
We're going to use "core promoter sequence-specific DNA binding".

PR #1216, opened 2020-08-19
Update most of the OntoFox imports

jamesaoverton:
I did not update 'core promoter binding' #1215 and left the `OBO_imports.owl` unchanged from #1209.


jamesaoverton:
Note: Almost all the difference are due to an updated OWLAPI version. I ran `robot diff` to check that the substantive changes to axioms looked good. The only one I'm a little worried about is NCBITaxon.

PR #1225, opened 2020-08-26
Add a base file to views

### LINKED ISSUE #1224: Create a "base" file
The base file should be created from the templates and not include any imports.

One caveat is that not all OBI terms are contained in templates, so we will need to make sure that these terms are still included in the base file.

beckyjackson:
Resolves #1224

The base file is created from obi-edit with imports removed, then all the modules created from templates merged back in.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,6 +130,16 @@ obi.owl: build/obi_merged.owl
 	--annotation owl:versionInfo "$(TODAY)" \
 	--output $@
 
+views/obi_base.owl: src/ontology/obi-edit.owl | build/robot.jar
+	$(ROBOT) remove --input $< \
+	--select imports \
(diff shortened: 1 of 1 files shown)


jamesaoverton (approved):
This may need revision, but it's good for now.

Please resolve conflicts, merge, and add a PURL entry for `obo/obi/obi-base.owl`.

beckyjackson:
Do we want to wait for a release to add the PURL entry or should I just add the non-dated version of the URL (without the release tag)? All the other "exact" entries use the date tags, e.g.: https://raw.githubusercontent.com/obi-ontology/obi/v2020-04-09/views/obi_core.owl

jamesaoverton:
I would like to start using an OBI base file, so I would like a PURL now. The PURL can point to `master` for now, but should use a dated version later.

Issue #1176, opened 2020-06-10
Specimen terms request 


linikujp:
Below is a list of verbatims copied from the EUA documents with limited redundant check. Those terms include synonyms, alternative terms and acronyms. Please sort it out with correct hierarchies before adding into CIDO. (Originally requested to CIDO https://github.com/CIDO-ontology/cido/issues/7)--- Per @bpeters42 request, I request the terms to be added into OBI. 

Please note the difference between  swab and specimen took using a swab. Swab is a device/tool, you may need to add a specimen for terms ending with swab.

anterior nasal swab
BAL fluid
BAL specimen
BALs
bronchial aspirates
bronchial washings
bronchial washings (BW)
bronchoalveolar lavage
bronchoalveolar lavages (BAL)/bronchial washings (BW)
lower respiratory specimens
lower respiratory tract aspirates
mid turbinate nasal swabs
midturbinate
mid-turbinate
mid-turbinate nasal swab
nasal
nasal aspirates
nasal swab
nasal wash
nasopharyngeal
nasopharyngeal (NP)
nasopharyngeal and oropharyngeal specimens
nasopharyngeal and oropharyngeal swab and bronchoalveolar lavage (BAL) specimen
nasopharyngeal aspirates
nasopharyngeal aspirates (NA)/nasal washings (NW)
nasopharyngeal awash/aspirate or nasal aspirates as well as sputum specimens
nasopharyngeal swab
nasopharyngeal swabs/wash
nasopharyngeal wash
nasopharyngeal wash/aspirates
NP/OP
oropharyngeal (OP)
oropharyngeal (throat) swab
oropharyngeal swabs/throat swabs
saliva specimens
sputum
throat (oropharyngeal) swab
tracheal aspirates
tracheal secretions
upper respiratory specimens
stool specimen
human serum or plasma (EDTA) specimen

linikujp:
Hi, I compared existing OBI terms and my list. 
1. The major terms that are lacking in OBI are: tracheal aspirate specimen, mid-turbinate nasal swab specimen. 
2. Other major issues will be to differentiate the wash and aspirates.
3. Hierarchical terms such as upper respiratory specimen and lower respiratory specimen.  

Will continue studying those, and submit a final excel sheet for new terms submission.

linikujp:
Dear everyone,

Can you recommend any existing relation linking a specimen to the anatomical part where the spwcimen is supposed to be collected? 

Thanks,
Asiyah

[...]

PR #1231, opened 2020-09-07
New terms, issue #1176

apmody:
New terms from issue #1176 

Calling `make all` in following error that needs to be resolved:

MANCHESTER PARSE ERROR the expression 'specimen and (is_specified_output_of some 'collecting specimen from organism') and ('derives from' some trachea)' at row 70, column 8 in table "src/ontology/templates/biobank-specimens.tsv" cannot be parsed: encountered unknown 'trachea'


jamesaoverton:
So far so good.

1. Please use the UBERON "replacement" terms from this comment: <https://github.com/obi-ontology/obi/issues/1176#issuecomment-689071343> When that's done, I'll update the UBERON imports to fix the parse error.

2. Please remove 'aspiration specimen collection' from the table. It's not a type of specimen, so we'll put it somewhere else.

jamesaoverton:
Thanks. I'll take care of the next step.

It would have been better not to commit the deletion of `biobank-specimens.owl`. I'll rebuild it.

jamesaoverton:
I updated the imports and made some changes to the table.

Please use Protege to add 'aspiration specimen collection' to `src/ontology/obi-edit.owl`.

There are still some problems to fix for the aspirate stuff.

beckyjackson:
@apmody there are a bunch of changes to [biobank-specimens.tsv](https://github.com/obi-ontology/obi/pull/1231/files#diff-058f793f62e0b37df89611d421ba14a1) that are making it difficult to see what's changed. Is there a reason quotes were added to the definition source field? Since these are TSVs, not CSVs, I don't think we need to enclose these in quotes.

Side note, there's also some trailing space on some of those values (e.g. `"Chris Stoeckert, NCI BBRB "`)

jamesaoverton:
I'm pretty sure that I ran `make sort` and it added these quotes (because the fields contain commas).

beckyjackson:
Do we want to keep them or should I clean that up?

jamesaoverton:
I think it's right now, and will be stable going forward.

src/ontology/templates/biobank-specimens.tsv
```diff
@@ -1,72 +1,79 @@
...
+OBI:0002611	vagina swab specimen		A specimen that is collected with a swab from the surface of a vagina.	"Chris Stoeckert, OBIB "	equivalent		vagina	UBERON_0000996	Chris Stoeckert	IAO:0000122
+OBI:0002612	fecal swab specimen	stool swab specimen	A specimen that is collected with a swab from the surface of a feces.	"Chris Stoeckert, OBIB "	equivalent		feces	UBERON_0001988	Chris Stoeckert	IAO:0000122
+OBI:0002613	environmental swab specimen		A specimen that is collected with a swab from the surface of an environmenal material.	"Chris Stoeckert, OBIB "	equivalent		environmental material	ENVO_00010483	Chris Stoeckert	IAO:0000122
 OBI:0002746	induced sputum specimen		A sputum specimen that is collected from an organism following its inhalation of a nebulized salt solution.	PMC3297553	subclass	sputum		UBERON_0007311	John Judkins	
+OBI:0002778	tracheal aspirate specimen	tracheal wash specimen	a specimen that is collected from the tracheal anatomic part (the tube connecting larynx to primary bronchi) of the respiratory system of an organism using a suction motion	"combined from wiki, pubmed, and dictionary"	subclass		trachea	UBERON_0003126	Asiyah Yu Lin|ORCID:0000-0003-2620-0345
```
beckyjackson, line 70: I know the term requester submitted these definitions & sources so you can replace the source if you rewrite the defintion. It looks like she didn't provide any specifics. Here are just some notes on improving this entry:

This definition can be simplified to match other definitions in this sheet (I didn't do a ton of research into this, so this might not be fully accurate):
> A specimen that is collected with suction from the trachea.

For definition sources, please use a pipe separator (`|`) instead of comma-separation. There should be no spaces between entries. See examples starting at line 77.

Also for definition sources, please provide specifics. If you are using an article from PubMed, include the ID (e.g., `PMID:1234`). For Wikipedia, just include the whole link. If the dictionary was an online dictionary, also include the link to the page that you got the detail from. The whole defintion source may end up looking like:
```
PMID:1234|http://en.wikipedia.org/wiki/Article_name|http://www.dictionary.com/term
```

src/ontology/templates/biobank-specimens.tsv
```diff
@@ -1,72 +1,79 @@
...
 OBI:0002746	induced sputum specimen		A sputum specimen that is collected from an organism following its inhalation of a nebulized salt solution.	PMC3297553	subclass	sputum		UBERON_0007311	John Judkins	
+OBI:0002778	tracheal aspirate specimen	tracheal wash specimen	a specimen that is collected from the tracheal anatomic part (the tube connecting larynx to primary bronchi) of the respiratory system of an organism using a suction motion	"combined from wiki, pubmed, and dictionary"	subclass		trachea	UBERON_0003126	Asiyah Yu Lin|ORCID:0000-0003-2620-0345	
+OBI:0002779	mid-turbinate nasal swab specimen		"a nasal swab specimen that is collected from the mid-turbinate area of  by a mid-turbinate nasal swab, which  guides the collector to the proper depth in order to ensure that the proper depth is reached when collecting sample from mid-turbinate area of the nose."	https://vimeo.com/135884130	equivalent		middle nasal concha	UBERON_0005921	Asiyah Yu Lin|ORCID:0000-0003-2620-0345
```
beckyjackson, line 71: > A specimen that is collected with a mid-turbinate nasal swab from the turbinate area of the nasal cavity.

I used: https://seattlechildrenslab.testcatalog.org/catalogs/185/files/11633

beckyjackson (changes requested):
The definitions need to be cleaned up a bit. I gave some examples on the first two, but mainly I think it would be good for these definitions to match the general pattern of this sheet:
> A specimen that is collected with X from Y

Where `X` is the method of collection (e.g., swab) and `Y` is the anatomical area (e.g. cervix)

Also make sure all the definitions start with a capital "A".

I also think it would be good to have definition sources for them. See my comment about sources on line 70.

beckyjackson:
@apmody please disregard the review, I will update the definitions. Thanks!

beckyjackson:
After updating the definitions, I re-ran `make sort` and it removed those quotes. I think that should be fine going forward?

beckyjackson (approved):

PR #1245, opened 2020-10-21
Fix 'antigen binding' placement

### LINKED ISSUE #1228: Term "Antigen binding" subclass of "molecular function"?
Hi, 

I was checking the OBI/Process class. I am a bit confused about the position of Term __Antigen binding__ http://purl.obolibrary.org/obo/GO_0003823. 

According to GO, it should be a subclass of Term __Molecular function__ http://purl.obolibrary.org/obo/GO_0003674  rather than a sibling of __Molecular function__. 

I am not sure if this is designed on purpose or should the GO hierarchy be used here?

Thanks for your reply.

Best,
Fuqi

beckyjackson:
Resolves #1228

@jamesaoverton @bpeters42 "core promoter binding" (GO:0001047) has been deprecated by GO and replaced by "core promoter sequence-specific DNA binding" (GO:0001046). Should I go ahead and update this or does this need to be discussed first? (This is what's causing the test to fail)


jamesaoverton:
We have #1215 for that, which I forgot about. Let's ask @bpeters42 and @rvita on our call today.

beckyjackson:
@jamesaoverton this is ready to merge. I replaced "core promoter binding" and updated the modules.

jamesaoverton (approved):

Issue #1250, opened 2020-10-30
NTR: endotracheal aspirate specimen & endotracheal tube specimen


johnwjudkins:
OBI:0002971
Label: endotracheal tube specimen
Definition: A specimen that derives from a biofilm that forms on the inner surface of an endotracheal tube while the tube is in the trachea. The specimen is collected upon removal of the tube from the trachea.

OBI:0002972
Label: endotracheal aspirate specimen: 
Definition: A specimen that derives from a biofilm that forms on the inner surface of an endotracheal tube while the tube is in the trachea. The specimen is collected by suction without removing the tube.

(this definition edited for clarity and consistency w/other specimen definitions)

Source for each: https://bmcpulmmed.biomedcentral.com/articles/10.1186/s12890-019-0926-3

According to https://en.wikipedia.org/wiki/Tracheal_tube, an endotracheal tube is a type of tracheal tube, so the term I'm proposing is different from the 'tracheal aspirate specimen' term that was also proposed.

johnwjudkins:
To provide an axiom for these terms consistent with other terms in the template, I want to add 'respiratory system fluid/secretion' (as child of 'organism substance') from UBERON to serve as the "gross anatomical part". Once the endotracheal tube is inserted, respiratory fluid containing bacteria forms on the endotracheal tube surface, allowing further bacterial growth (http://rc.rcjournal.com/content/60/1/21). So either of these specimen types derive from respiratory fluid.

PR #1256, opened 2020-11-18
endotracheal aspirate specimen & endotracheal tube specimen

johnwjudkins:
#1250


johnwjudkins:
Closing without merging because 1) this has become messy 2) I have better axioms for the specimens and will submit a different request.

PR #1261, opened 2020-12-10
endotracheal specimens

johnwjudkins:
#1250 (Apparently I can only add the UBERON term now since the specimen terms depend on it; that's why I reverted the commit)


jamesaoverton:
@johnwjudkins I restored the template rows and updated the OWL modules. Can you please double check?

johnwjudkins:
Looks great - thanks @jamesaoverton !

Issue #1217, opened 2020-08-19
NTRs for arthropod specimen collection


johnwjudkins:
The VEuPath team has curated the following terms relating to the capture of insects (and other arthropods) for study, and we think they are a good fit for OBI. They can also be viewed in our latest release. We classify each term as a descendant of either 'device' or 'planned process'.

Label | Definition | Parent
-- | -- | --
arthropod specimen collection by aspiration | A live arthropod specimen collection process that involves the use of an aspirator, a device which draws an arthropod by suction into a container. | adult arthropod specimen collection process
arthropod resting outdoors on artificial support specimen collection by aspiration | An arthropod resting outdoors specimen collection by aspiration that occurs while the arthropod rests on a processed material such as a wall or furniture. | arthropod resting outdoors specimen collection by aspiration
arthropod resting outdoors on natural support specimen collection by aspiration | An arthropod resting outdoors specimen collection by aspiration that occurs while the arthropod rests on an unprocessed material such as a plant. | arthropod resting outdoors specimen collection by aspiration
animal-biting arthropod specimen collection by aspiration | An arthropod specimen collection by aspiration that occurs while the arthropod is biting an animal. | arthropod specimen collection by aspiration
animal-biting outdoor arthropod specimen collection by aspiration | An animal-biting arthropod specimen collection by aspiration that occurs outside a building. | arthropod specimen collection by aspiration
arthropod resting in animal house specimen collection by aspiration | An arthropod specimen collection by aspiration that catches specimens resting inside an animal house. | arthropod specimen collection by aspiration
arthropod resting outdoors specimen collection by aspiration | An animal-biting arthropod specimen collection by aspiration that occurs while the arthropod is resting outside a building during a period of inactivity. | arthropod specimen collection by aspiration
knockdown arthropod specimen collection process | An arthropod specimen collection process that involves exposing arthropods to pesticide, then collecting them once they become dead or immobilized. | arthropod specimen collection process
live arthropod specimen collection process | An arthropod specimen collection process meant to collect live specimens. | arthropod specimen collection process
artificial pit shelter | An arthropod trap, consisting of a pit dug in the ground to provide a space for arthropods to rest and be collected. | resting arthropod trap
baited net trap | An arthropod trap designed to work with an animal (which could be human) inside a netted structure. This structure is designed so that an arthropod, attracted by the animal, can fly into the netted structure but can neither escape nor reach the animal. | arthropod trap
[...]

PR #1278, opened 2021-01-11
Column in sample-collection template for specimen type (output)

johnwjudkins:
Mentioned in #1217. This would add a column to create an axiom connecting the specimen collection process to the bearer of the specimen role created in the process.

`C has_specified_output some (% and 'has role' some 'specimen role')`


jamesaoverton:
Makes sense to me.

Issue #1232, opened 2020-09-11
Does OBI want the following sequence assembly datums?


ddooley:
We are deciding where to submit to.  I don't thing SO has anything in the datum department.  If not OBI then we'll put in GenEpiO.

d.

SUBCLASS OF | LABEL | DEFINITION | DEF SOURCE | example of use | Part Of | Exact Synonym@en
-- | -- | -- | -- | -- | -- | --
lab sequencing datum | amino acid sequence start site (user submitted) | The position of the first amino acid in a user-submitted protein (query) sequence. | Emma |   |   | Query Start (aa)
lab sequencing datum | amino acid sequence stop site (user submitted) | The position of the last amino acid in a user-submitted protein (query) sequence. | Emma |   |   | Query Stop (aa)
lab sequencing datum | nucleic acid sequence start site (user submitted) | The position of the first nucleotide in a user-submitted nucleotide (query) sequence. | Emma |   |   | Query Start (nt)
lab sequencing datum | nucleic acid sequence stop site (user submitted) | The position of the last nucelotide in a user-submitted nucleotide (query) sequence. | Emma |   |   | Query Stop (nt)
lab sequencing datum | amino acid sequence start site (reference) | The position of the first amino acid in a protein reference (subject) sequence. | Emma |   |   | Subject Start (aa)
lab sequencing datum | amino acid sequence stop site (reference) | The position of the last amino acid in a protein reference (subject) sequence. | Emma |   |   | Subject Stop (aa)
lab sequencing datum | nucleic acid sequence start site (reference) | The position of the first nucleotide in a nucelotide reference (subject) sequence. | Emma |   |   | Subject Start (nt)
lab sequencing datum | nucleic acid sequence stop site (reference) | The position of the last nucelotide in a nucelotide reference (subject) sequence. | Emma |   |   | Subject Stop (nt)

ddooley:
I'll double check with the SO sequence ontology folks about taking these in, and when closer to finalization, with the protein ontology folks about labels/definitions.  Regardless, Emma could you elaborate on the terms "user submitted" and "reference".  

Part of the challenge in defining these is that an agency's "user submitted" sequence to some repository is at the very same time their own internal reference sequence, especially if they are providing them as a database.  

[...]

PR #1279, opened 2021-01-13
This is a set of 8 protein/gene sequence start/stop terms

ddooley:
These terms awaiting any final review/modification/comment.


jamesaoverton:
This builds and passes all the tests.

jamesaoverton:
See https://github.com/obi-ontology/obi/issues/1232

I will remove the extra spaces around the splits: ` | `.

PR #1281, opened 2021-01-15
Fix links to view/ in readme.md

cbielow:
when looking for `obi.obo`, I found it mentioned in the readme.md, but the link was broken.
This should fix it.


jamesaoverton:
Much appreciated!

Issue #1247, opened 2020-10-26
"nasal cavity" (UBERON:0001707) has no parent


beckyjackson:
"nasal cavity" was added to the UBERON import, but it is an "immaterial anatomical entity" and we do not have this as an upper-level term for the UBERON import.

I think there are three options:
1. Add "immaterial anatomical entity" (or "anatomical cavity") as an upper level term
2. Find a new home for "nasal cavity"
3. Replace "nasal cavity" with some material anatomical entity (upon a quick search, I wasn't able to find a good replacement, but I also am not well-versed in the anatomy of the nose 🙂 )

bpeters42:
where are we using it? For nasal swab I presume? That means we should go with 3), and refer to 'content of nasal cavity'. I have had discussions with Chris Mungall that we should create sibling classes for all the 'immmaterial entities' to link to 'content of'

beckyjackson:
It's just for "nasal aspirate specimen".

Would "content of nasal cavity" be a new OBI term? I guess we could actually map it to "nasal cavity", though, because that would require importing "nasal cavity" anyway...

bpeters42:
In my order of happiness: 

a) This to be solved in the Uberon layer, and I would request for each Uberon 'cavity' entry a sibling 'material entity located in cavity' entry. Any suggestion @cmungall has on how realistic that is to get done any time soon would be helpful. It seems like there is agreement on this, but this is touching on very old issues, so it might not happen. I naively think though that this specific proposal (creating separate classes for cavity / content of) has the potential to be least disruptive? 
b) change the design pattern in OBI for these specimens to be sample_of some (located_in some anatomical cavity) 
c) create an OBI placeholder class, to be removed  in the future, of 'material entity and located in cavity

I would love a); b) is realistic and ok; c) is cruel and listed for completeness 

I prefer

beckyjackson:
Here's an option d)

We found "nasal cavity mucosa" in UBERON, which is a material anatomical entity. If we replace "nasal cavity" with this, we won't need to change anything else.

PR #1285, opened 2021-01-15
OntoFox inputs for new assays from TURBO, issue #1284

turbomam:


jamesaoverton:
These OntoFox input files are fine, although I don't understand why each one now has a line with just `#`.

The hard part will be actually running OntoFox and updating the OWL files, because recently we've had all sorts of trouble with unwanted annotations from all over the place.

turbomam:
@jamesaoverton would you have a few minutes to chat in the next couple of days?

The nothing-but-# lines are a reminder of where my additions started. I was expecting that I might have use OntoFox's ability to reassert the parent class on some of them, since that has been done pretty extensively for existing imports. Does that have anything to do with what you've been seeing with unwanted annotations?

jamesaoverton:
You deleted `src/ontology/modules/epitope-assays.owl`.

src/ontology/OntoFox_outputs/OBO_imports.owl
```diff
@@ -18,6 +18,92 @@
...
+    <!-- http://purl.obolibrary.org/obo/IAO_0000111 -->
+
+    <owl:AnnotationProperty rdf:about="http://purl.obolibrary.org/obo/IAO_0000111">
+        <obo:IAO_0000111>editor preferred label</obo:IAO_0000111>
+        <obo:IAO_0000111 xml:lang="en">editor preferred label</obo:IAO_0000111>
+        <obo:IAO_0000111>editor preferred term</obo:IAO_0000111>
+        <obo:IAO_0000111 xml:lang="en">editor preferred term</obo:IAO_0000111>
+        <obo:IAO_0000111>editor preferred term~editor preferred label</obo:IAO_0000111>
+        <obo:IAO_0000111 xml:lang="en">editor preferred term~editor preferred label</obo:IAO_0000111>
+        <obo:IAO_0000111>编辑首选术语</obo:IAO_0000111>
```
jamesaoverton, line 34: These are the sorts of annotations that we've been getting from OntoFox, but I really don't want them in OBI.

jamesaoverton:
`src/ontology/modules/biobank-specimens.owl` has changed without corresponding changes to `src/ontology/templates/biobank-specimens.owl`.

At a glance, the OntoFox imports look OK except for the `OBO_imports.owl`. Every time we update it, it's a mess. I've been cleaning it up manually by just picking the changes that I want to commit, but it's a big pain.

src/ontology/OntoFox_outputs/OBO_imports.owl
turbomam: I agree and am glad to help manually or programmatically if you want.

I agree that `src/ontology/modules/epitope-assays.owl` is missing, but I attributed that to the fact that the make processes first cat '' to the module OWL files, then try to rebuild them with `robot template`.  'Oncostatin M production' is required by  row 403 in `epitope-assays.tsv`, but it couldn't be found in `obi-edit.owl`, so that task failed. 

I haven't found anything that I can do that that point to get the module templating task to work again, because they all seem to depend on one another, in terms of being imported into `obi-edit.owl`, and therefore serving as `--input`s into each other's build.

I may have gotten some of that wrong and am not really complaining. If you think you know what I'm getting at, please set me straight when you can.

----

> echo '' > src/ontology/modules/epitope-assays.owl
java -jar build/robot.jar --prefix "REO: http://purl.obolibrary.org/obo/REO_" merge \
	--input src/ontology/obi-edit.owl \
	template \
	--template src/ontology/templates/epitope-assays.tsv \
	annotate \
	--ontology-iri "http://purl.obolibrary.org/obo/obi/dev/modules/epitope-assays.owl" \
	--output src/ontology/modules/epitope-assays.owl
MANCHESTER PARSE ERROR the expression 'has_specified_output some ('information content entity' and ('is about' some 'Oncostatin M production'))' at row 403, column 11 in table "src/ontology/templates/epitope-assays.tsv" cannot be parsed: encountered unknown 'Oncostatin M production'
For details see: http://robot.obolibrary.org/template#manchester-parse-error
Use the -vvv option to show the stack trace.
Use the --help option to see usage information.
make: *** [src/ontology/modules/epitope-assays.owl] Error 1
make: *** Deleting file `src/ontology/modules/epitope-assays.owl'

jamesaoverton:
'Oncostatin M production' is http://purl.obolibrary.org/obo/GO_0044808 and the label was changed in your updated import, which broke the template. In hindsight, it would have been better to update all the OntoFox import before adding new terms, which would have isolated that problem.

It's a big PR, so there's going to be problems like this. Maybe we should schedule a call to work through it together?

turbomam:
Yes, thanks, let's do a call. How about Thursday morning, any time Friday, or any time Tuesday the 26th?

jamesaoverton (approved):

cstoeckert (changes requested):
Two issues found with these changes to be fixed:
1. chebi imports.owl - rac-lactic acid not under material entity. not under anything in obi-dev
2. obo imports.owl - new PR terms not under anything: 'cytosolic creatine kinase complex, MB-type', 'hemoglobin A1c', 'lactate dehydrogenase complex'
Two additional problems found but not due to new assays.
1. http://purl.obolibrary.org/obo/GO_0002390 platelet activating production is deprecated in GO and not imported. Doesn’t appear to be used in any axioms in current OBI release.
2. nasal cavity not under anything in obi-dev: problem in current release!

turbomam:
Per the feedback from @cstoeckert and assignment from @jamesaoverton , I will reroot the terms @cstoeckert mentioned with the deepest relevant terms already ion OBI. I'm going to do that for 'nasal cavity' too, but I have a sense that somebody else may have added it. 

UBERON asserts the parent as [anatomical cavity](http://purl.obolibrary.org/obo/UBERON_0002553). I will reroot it under [immaterial entity](http://purl.obolibrary.org/obo/BFO_0000141).

I will also comment out [platelet activating production](http://purl.obolibrary.org/obo/GO_0002390)



Go to external page  Copy

jamesaoverton:
Regarding "nasal cavity", let's replace it with "nasal cavity mucosa" as in #1247

turbomam:
@cstoeckert I added nasal mucosa but did not remove nasal cavity because it is a requirement for some existing axioms.
Do you know how to indicate that your requested changes have been satisfied?

cstoeckert:
@turbomam you need to request a re-review (which you have kind of done). I'll review and update. Thanks

cstoeckert (approved):
Identified terms now have proper parents.
Note that while this fixes #1247 still need to adjust the axiom for nasal swab specimen bu that can be a different PR.

jamesaoverton:
I completed the fix for #1247.

Issue #1263, opened 2020-12-11
Definition and axioms of 'scalar measurement datum'


APCox:
The current definition of 'scalar measurement datum' is: "a measurement datum that is composed of two parts, numerals and a unit label." This means that measurement datum items that utilize a scale without units (e.g. a Likert or other rating scale) must be typed as 'measurement datum' rather than 'scalar measurement datum'. Is this the intended usage? If so, then it would be useful to add a term such as 'ordinal measurement datum' as a subtype of 'measurement datum' (and possibly the parent of 'scalar measurement datum').

More importantly, there appears to be an unnecessary duplication of entities caused by the axiomatization of 'scalar measurement datum'. Specifically, the axioms for 'scalar measurement datum' include:
1. 'has measurement unit label' min 1 owl:Thing
2. 'has measurement value' min 1 rdfs:Literal
3. 'has value specification' some 'scalar value specification'

These axioms appear unproblematic until you notice that 'scalar value specification' includes the axioms:
1. 'has measurement unit label' min 1 owl:Thing
2. 'has specified numeric value' min 1 rdfs:Literal

such that the value and unit label are duplicated for both the scalar measurement datum and the scalar value specification. The result is 2 presumably distinct ICEs -- each of which has its own value and unit label. It is unclear to me what value is added by including axiom (3) given the current axioms. As far as I can tell, (3) currently creates an extra ICE with duplicate values. Clarification would be greatly appreciated.

ddooley:
Thanks for raising this problem too - it needs resolution and OBI is committing to a solution before the end of January.  This duplication dates back to when OBI introduced value specifications as a second and improved way to capture the value associated with a datum (and separate scales of measure, so in a data structure allowing both a Celsius and Fahrenheit value on a temperature measurement datum); previous to that datums could have 'has measurement value' and 'has measurement unit label' data properties directly as you note.  

The unfinished business of that work was to drop these data properties from 'scalar measurement datum'.
I suggest we do that now.

An RDF data structure involving 'has unit' and 'has numeric value' properties (common in the non-OBI world), and OBI datum having 'has measurement unit label' and 'has measurement value' could be converted into a value specification unit and value for data compatibility and OWL driven reasoning.

APCox:
[...]

PR #1299, opened 2021-02-19
scalar measurement datum adjustment, issue 1263

ddooley:
Removing 'has measurement unit label' min 1 Thing
Removing 'has measurement value' min 1 rdfs:Literal
per issue https://github.com/obi-ontology/obi/issues/1263
These axioms are archaic, they were outdated with introduction of axiom: 'has value specification' some 'scalar value specification'


ddooley:
Additionally,  a 'scalar measurement datum' has a 'scalar value specification' with a  'has measurement unit label' axiom.  We are moving to relax this, barring any objections. Not all scalars have units.

jamesaoverton:
Can you please make these changes in `src/ontology/` (probably `obi-edit.owl`)?

The `obi.owl` file is automatically generated, so these changes will be lost on rebuild. Also they won't be reflected in any of the other generated files, such as OBI Core.

ddooley:
Argh, thank's for catching that. I edited a copy I had open without even thinking of checking.  Will do.

ddooley:
Revised in separate pull requests to obi-edit.owl and iao-edit.owl (#241)

alanruttenberg:
It is uncool to change the meaning of an existing term. Please define a new term instead.

ddooley:
Hi Alan. So I'm almost always for that, and some suggestions are happening in that regard.  This particular "scalar measurement datum" issue is dealing with a logical/data structure conundrum - that "scalar measurement datum" currently has a value specification AND direct measurement value/unit in its axioms.  I think the problem was introduced the moment value specification was added to it years ago, and never cleared up.  Do we leave these conflating axioms as is, and deprecate "scalar measurement datum" since it is practically unusable with them in place?  

Personally I am ok with deprecating "scalar measurement datum" and creating cleaner replacements: a unitless "scalar measurement data item" and "unit-bearing scalar measurement data item". 

But there remains the issue of preferring "value specification" over direct attachment of measurement value/unit to a data item. Where do you stand on that?  

The upcoming COB of course seems to be offering an alternative with its "has quantity" which is a third and apparently simpler construct from the eyes of graph db folks as it incorporates both value and optionally user defined unit.

alanruttenberg:
Hi, 

Thanks for a clear description of the issue. I concur that the problem originates with the addition of value specification. It's at that point that it would have been better to deprecate. I'm wondering if a property chain could be used to populate the unit from the value specification.

However, it should be noted that the SI specifies that dimensionless quantities have a unit: 1. Moreover this [NIST paper](https://iopscience.iop.org/article/10.1088/0026-1394/52/1/40) suggests that such dimensionless aren't always coherent, and suggests additional "units" to replace 1 are needed.  Even considering ratios, if the quantities have no dimension, that suggests that dimensionless quantities can be combined, for instance added, and that's not the case.

In my view, It's unfortunate that the replacement value specification includes no constraints at all. 

I don't understand what the COB is saying about has quantity. FWIW I'm a 'graph db' folk - probably used it before any others in OBO. Still, I don't make ontological decisions based on an implementation of a store. The recommendation to use a user-defined datatype will mean that the OWL reasoners will error out or ignore them.  

I'm not sure what to do here. It's a mess.

PR #1301, opened 2021-02-23
Removing a unit label from scalar value spec.

ddooley:
Not all scalars have units, so we are removing a scalar value specification's 'has measurement unit label' axiom. Addresses https://github.com/obi-ontology/obi/issues/1263


ddooley:
Also see sister pull request for IAO: https://github.com/information-artifact-ontology/IAO/pull/241

jamesaoverton:
@ddooley I'm going to mark this as a Draft while we build consensus.

Issue #1303, opened 2021-03-01
enzymatic activity assay NTR


turbomam:
Label: Enzymatic activity level assay

Definition: an assay in which the activity (not abundance) of any enzyme is determined? measured?

Examples: ALT
vs measuring ATP levels with luciferase, in which an enzyme is a reagent

Parent class: assay

see also #1304

cstoeckert:
Propose going for a very general definition:
"An assay in which the activity of an enzyme is determined."

jamesaoverton:
Discussed with the CEBS team on 2022-04-20. This is our suggestion:

Label: enzymatic activity assay
Definition: An assay in which the activity of an enzyme is determined.
Parent class: assay
Target Entity: catalytic activity (for lack of a more specific 'enzyme activity')

mgiglio99:
Discussed on April 25th call.
Decided to make slight changes.
label to: enzymatic activity level assay
definition to: An assay in which the level of catalytic activity of an enzyme is determined.

jamesaoverton:
In reviewing PR #1572 we noticed that 'insecticide resistance by detecting enzyme activity assay' is somehow related. These could maybe use the new 'enzyme activity level assay' as a detection technique. Not a high priority.

PR #1313, opened 2021-03-15
enzyme activity assay #1303

turbomam:
just added the class, so that others can start adding subclasses


jamesaoverton:
The definition currently ends with a question mark, so more work needs to be done here.

johnwjudkins:
https://github.com/obi-ontology/obi/issues/1303

beckyjackson:
Not sure what the status on this is, but I think the label needs to be lowercase. Is that correct?

bpeters42:
These terms are useful. But current label and definition are problematic. There are also some discussions (Jennifer F with @jamesaoverton) on the logical modeling on how to capture 'enzymes' correctly.

jamesaoverton:
Closed by #1572

PR #1314, opened 2021-03-16
fixed analyte to be parathyroid hormone

cstoeckert:
analyte was wrong so corrected and imported parathyroid hormone from PRO. updated OBO_imports with Ontofox


jamesaoverton:
Discussed on 2021-03-29 call: The OBO_output.owl has problems. I will fix them manually this time and merge. Then I will try to fix the larger problem by breaking the imports into smaller pieces.

PR #1315, opened 2021-03-16
rearrange Header columns in epitope-assays

hectorguzor:
Rearranged Header columns in epitope-assays.tsv to better match assays.tsv. Does not change the content of epitope-assays.owl.


jamesaoverton:
I confirmed that this only changes the template file, not the generated OWL, as desired.

PR #1321, opened 2021-03-29
Split OBO imports into individual modules

beckyjackson:
This PR removes `OBO_imports` and instead splits that file into each individual module.


jamesaoverton:
I checked this pretty thoroughly, and it seems to be working perfectly. Thanks!

PR #1323, opened 2021-03-30
epitope-assays.tsv columns cleanup

hectorguzor:
This removes "Bcell process result" and "Tcell process result" columns, and replaces "Biological process" column with "Target entity" in epitope-assays.tsv.


jamesaoverton (approved):
Hector and I both checked this over pretty carefully. It does change some of the logic, simplifying things, but it keeps the tree structure the same.

PR #1324, opened 2021-03-31
WIP: Import workflow update

beckyjackson:
This switches the import workflow to use [Gizmos](https://github.com/ontodev/gizmos).

This is using a version of `gizmos` in a branch here: https://github.com/ontodev/gizmos/pull/79


src/ontology/imports/chebi_imports.owl
```diff
@@ -74,7 +57,6 @@
         <rdfs:subClassOf rdf:resource="http://purl.obolibrary.org/obo/CHEBI_23367"/>
         <obo:IAO_0000111 rdf:datatype="http://www.w3.org/2001/XMLSchema#string">water</obo:IAO_0000111>
         <obo:IAO_0000115 rdf:datatype="http://www.w3.org/2001/XMLSchema#string">An oxygen hydride consisting of an oxygen atom that is covalently bonded to two hydrogen atoms.</obo:IAO_0000115>
-        <obo:IAO_0000412 rdf:resource="http://purl.obolibrary.org/obo/chebi.owl"/>
```
jamesaoverton, line 77: This is the 'imported from' link, which we aren't currently adding with the new workflow. I kinda prefer to have these, because it's kinda useful if you're looking at `obi.owl`. If we do have these then the diff will be much smaller.

jamesaoverton:
For OBI developers wondering about this draft PR, it changes the import system from using the OntoFox web service to using locally installed files and tools. The key benefits are:

1. A cleaner import configuration in two tables [`imports.tsv`](https://github.com/obi-ontology/obi/blob/import-update/src/ontology/imports/imports.tsv) and [`config.tsv`](https://github.com/obi-ontology/obi/blob/import-update/src/ontology/imports/config.tsv). This will also enable better tools and interfaces for managing imports.
2. More control over which version of each ontology we import. When calling OntoFox we get whatever version they happen to have loaded in their triplestore, but with this system we can pick and choose exactly which OWL file to import from.
3. Efficient processing of large ontologies, such as NCBITaxon and Protein Ontology.

The tools we're using to do this are very new. This draft PR is really meant as a thorough test of the tools against OBI's pretty diverse imports. We won't merge it into OBI until (1) the tools are stable, and (2) have the approval of OBI developers.

PR #1327, opened 2021-04-01
Update clinical assays

jamesaoverton:
In this PR I'm trying to coordinate recently added clinical assays for TURBO with a use case from the National Toxicology Program (NTP). I've tried to break my changes into distinct steps for easy review:

1. lowercase 'lipid' in the labels for four 'lipid panel reflex direct venous blood ...' assays
2. normalize the use of single-quotation marks in assay.tsv (no change to ontology content)
3. add 'count' to the label of 68 cytometry assays: @cstoeckert says that these assays specifically result in absolute counts of cells, and the textual definitions all say "measures the count of X". However NTP needs similar cytometry assays that don't result in counts, but percentages or maybe relative abundance. So the labels of these 79 are too general, and adding 'count' to the label fixes this problem (I think).
4. fix textual definitions of 11 hematocrit assays: they said "measures the count of hematocrit", but now they say "measures the proportion of red blood cells"
5. change 14 'count assays' to be subclass not equivalent: This is needed to distinguish them from more general "abundance" assays that I want to add. Most of them were already subclass.
6. add 19 'abundance' assays that group various 'count' and 'venous blood' assays: NTP will either use or subclass these. We would prefer to add them now because they change the assay tree, as long as the TURBO team is OK with these changes.

ID | Label
---|---
OBI:0003009 | ferritin assay
OBI:0003010 | fibrinogen assay
OBI:0003011 | iron assay
OBI:0003012 | ionized calcium assay
OBI:0003013 | transferrin assay
OBI:0003014 | uric acid assay
OBI:0003015 | globulin assay
OBI:0003016 | myelocyte assay
OBI:0003017 | metamyelocyte assay
OBI:0003018 | hematocrit assay
OBI:0003019 | promyelocyte assay
OBI:0003021 | basophil assay
OBI:0003022 | neutrophil assay
OBI:0003023 | eosinophil assay
OBI:0003024 | lymphocyte assay
OBI:0003025 | monocyte assay
OBI:0003026 | reticulocyte assay
OBI:0003027 | macrophage assay
OBI:0003028 | platelet assay


jamesaoverton:
@cstoeckert Are you OK with the work so far, especially step 3?

cstoeckert (changes requested):
@jamesaoverton the changes you made are good except for a subset of cytometry assays involving 'hematocrit' where you added 'count'. The problem was caused by us - your change helped me identify it in the review. Hematocrits are proportion (percentages) not counts. We programmatically generated the definitions so ended up with "A cytometry assay that measures the count of hematocrits .." which does not make sense. Can we change the textual definition of the hematocrit assays to "A cytometry assay that measures the proportion of red blood cells.."? The original assay labels should be fine (i.e., don't add count).

cstoeckert (approved):
Thanks @jamesaoverton for the quick fix!

jamesaoverton:
Thanks @cstoeckert. I implemented that change to 11 hematocrit assays and updated the PR description.

Now I'll work on adding some new terms.

jamesaoverton:
@cstoeckert I edited the PR description again with more details. Can you please open `src/ontology/obi-edit.owl` in Protege, run HermiT, and take a look at these new terms in the inferred class view? I think they're straightforward groupings, but if you have any concerns just let me know. Thanks!

jamesaoverton:
Arg, I didn't actually submit my changes to the PR description until just now. Sorry.

cstoeckert (changes requested):
@jamesaoverton I think the analyte assays are good now. Thanks!
The cytometry assays need further work. 
1. cytometry assays includes measure of characteristics (not just abundance) so equivalence axiom would infer any assay that is about basophils (for example) to be a basophil assay. That doesn't match text definition of 'measure the abundance'
2. some text definitions have 'measure the count' instead of 'measure the abundance' such as metamyelocyte, myelocyte, promyelocyte. Should make these consistent.
One way to address these points is to make these general classes, truly general and make the text definition 'measure the abundance and/or characterstics' in place of 'measure the abundance' or 'measure the count'

jamesaoverton:
Good points @cstoeckert. I made those changes. Thanks, I really appreciate your help.

Now I'll turn off "Draft" mode, and plan to discuss this PR on Monday's call (briefly, I hope).

cstoeckert (approved):
Looks great! Thanks @jamesaoverton !

PR #1330, opened 2021-04-09
Standardize annotations in ROBOT templates

beckyjackson:
Over time, we have added more and more ROBOT templates, but we never specified a standard pattern for the elements to be included. Because of this, the templates have ended up with differing annotation properties in each, and the same annotation properties might have different headers between templates. While we expect the logical columns to differ between modules, terms across OBI should share the same annotations. Therefore, every ROBOT template should include each of these annotation properties.

This PR standardizes the ROBOT templates to share the same annotation properties:
- ontology ID
- label
- editor preferred term
- has curation status
- alternative term
- defintion
- defintion source
- example of usage
- term editor
- editor note
- curator note
- ontology term requester
- term tracker item
- logical type (not an AP, but still shared)

All headers are standardized to be lowercase as well.

The changes to the modules are just from missing `SPLIT=|` statements in the ROBOT template strings, as well as a couple of quoting fixes. No content in the templates was changed.


jamesaoverton:
Discussed 2021-05-03: No objections to the general approach.

In theory this PR should not change the OWL. In fact there are changes to OWL.

- `devices.owl` The changes to quotation marks seem fine
- `obsolete.owl` The changes seem finel
- `biobank-specimens.owl` seem wrong to me. People have been using pipes and ORCIDs incorrectly, and enforcing the splits makes it worse. I think the smallest change would be to use "{name} {orcid}", which was the intention, by removing inappropriate pipes from the tables. I'll request this change. In the future we need a consistent way of referring to people with human-readable names and ORCIDs.

jamesaoverton (changes requested):
Some of the splits in `biobank-specimens.owl` are wrong because there are too many pipes in the tables. @beckyjackson please remove the extra pipes so that entries have "{name} {orcid}" format.

jamesaoverton (approved):

PR #1337, opened 2021-04-25
Add pooled clone seq assay Issue 1106

DanBerrios:


bpeters42:
The definition should refer to the parent class: 

A DNA sequencing assay in which the DNA input is derived from pooled clones such as BACs and Fosmids. 


--

bpeters42:
The 'make sort' should ideally be run prior to commit.

bpeters42 (changes requested):
See comments.

bpeters42 (approved):

DanBerrios:
FYI, I gave up on editing the tsv on Windows Excel..there was no way to export the file that didn't cause quoting (that I could find). I edited in NotePad++.  Does anyone know if OpenOffice Calc causes problems too?

jamesaoverton:
The `make sort` script is supposed to fix the quoting but I think it isn't. @beckyjackson make a new version here that fixes that: https://github.com/obi-ontology/obi/blob/standardize-templates/src/scripts/sort-templates.py

There's another problem for display on GitHub that I'm trying to fix.

beckyjackson:
Missing newline in `devices.tsv` showed up when I ran `make sort`, which is why that file still has a change.

PR #1340, opened 2021-04-27
Add locate script

### LINKED ISSUE #1339: Write script to say where a term is defined (and used?)
The OBI README explains that terms are defined in various different files: https://github.com/obi-ontology/obi#editing

It's not *tooooo* hard for a person to figure this out, but it would be nice if it was easier. Computers are dumb, so it would be nice to have some code that would simply answer this question. Things like `grep` are too general. The script needs to know a little bit about ROBOT templates and OWL files and OntoFox input files.

The minimum thing I'd like is this (the name `locate` is just a suggestion, tabs would be better than spaces but they're hard to type into GitHub comments):

```sh
$ src/script/locate OBI:0000070 CHMO:0000087 GO:0000785
OBI:0000070   src/ontology/obi.owl:3769
CHMO:0000087  src/ontology/assays.tsv:3
GO:0000785    obi/src/ontology/OntoFox_inputs/GO_input.txt:8
```

It would be amazing if it could also work with labels:

```sh
$ src/script/locate "assay" "fluorescence microscopy" "chromatin"
```

It would be even more amazing to get usage information, but I'd be happy with the basics.

beckyjackson:
Resolves #1339 - added to README:

### Finding Terms

To find where a term lives, you can use [`src/scripts/locate.py`](src/scripts/locate.py). This requires you first to build a database from the merged OBI file:
```
make build/obi_merged.db
```

Then you can run the script to find terms by ID or label by passing them as a space-separated list, for example:
```
src/scripts/locate.py OBI:0000070 CHMO:0000087 GO:0000785
```

Labels should be enclosed in double quotes:
```
src/scripts/locate.py "assay" "fluorescence microscopy" "chromatin"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,23 @@ build/robot.jar: | build
 ROBOT := java -jar build/robot.jar --prefix "REO: http://purl.obolibrary.org/obo/REO_"
 
 
+### RDFTab
+#
+# Use RDFTab to create SQLite databases from OWL files.
+UNAME := $(shell uname)
+ifeq ($(UNAME), Darwin)
+	RDFTAB_URL := https://github.com/ontodev/rdftab.rs/releases/download/v0.1.1/rdftab-x86_64-apple-darwin
+	SED = sed -i.bak
+else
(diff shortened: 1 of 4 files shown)


jamesaoverton:
Awesome, thanks!

PR #1342, opened 2021-04-29
Fix devices template

beckyjackson:
The `devices.tsv` template has two extra columns with no headers currently. This is an issue for trying to process this file programmatically and also causes issues with GitHub's table display.

The file is now fixed to display correctly: https://github.com/obi-ontology/obi/blob/e647433c28f4dd5e4be73b5b40f3de055ee5be76/src/ontology/templates/devices.tsv


jamesaoverton (approved):

jamesaoverton:
@ddooley There was a column without a header that had some notes. I think you probably added those. Becky moved them into "editor notes". We're going to merge this now, but if you don't want them there, we'll delete them later.

ddooley:
Oh, I thought robot would just ignore that but now I see that still gets in the way of nice Github display.  There is no need for including that note column - accident from direct cut and paste from another source table.

jamesaoverton:
Ok. @beckyjackson please delete these notes when you get a chance.

PR #1344, opened 2021-04-30
Add script to relabel terms

beckyjackson:
From the docs (requires you to first `make obi-dbs`)

### Relabeling Terms

Since labels are used in templates, manually finding and replacing all usages of a term can be difficult and tedious. Instead, you can use [`src/scripts/relabel.py`](src/scripts/relabel.py) to automatically update a term's label and its usages.

You can run this by passing the term you want to update and the new label:
```
src/scripts/relabel.py CHMO:0000087 "microscopy with fluorescence"
```

You can also pass the old label as the first argument, as long as it is enclosed in double quotes:
```
src/scripts/relabel.py "fluorescence microscopy" "microscopy with fluorescence"
```

Make sure to commit all changed files to ensure that all usages are updated.


jamesaoverton:
This may not work perfectly, but actually using it is the best way to find problems.

PR #1351, opened 2021-05-13
Add workflow to make OBI Excel sheet for editing templates

beckyjackson:
This workflow allows editors to create an Excel spreadsheet from our templates, edit it, and convert the spreadsheet back to TSVs to build the modules:
1. Run `make obi.xlsx`
2. Edit `obi.xlsx`
3. Run `make update-tsv`
4. Run `make modules`

Note that this requires `openpyxl` for the Python scripts. Install with:
```
python3 -m pip install -r requirements.txt
```

... or:
```
python3 -m pip install openpyxl
```

(Since we just have one requirement, maybe we don't need a requirements file? I don't know.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 build/
 catalog-v001.xml
 junk/
+*.xlsx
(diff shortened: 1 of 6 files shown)


jamesaoverton (approved):
Worked for me.

I don't love the instructions for global Python install, but venv instructions are worse.

PR #1352, opened 2021-05-14
adds 'measurand' column to assays.tsv (issue #1062)

### LINKED ISSUE #1062: In assays.tsv, "analyte" column should be "measurand" column
The OBI ontology distinguishes between analyte and measurand roles; however, the assays.tsv template doesn't reflect that recent change - the template has an "analyte" column, but one of its cell contains the name of a virus (a virus is not a molecular entity and so can't bear an analyte role). So I propose changing the "analyte" column to the "measurand" column.

hectorguzor:
Adds 'measurand' column to `assays.tsv` and moves two values from 'analyte' column into it. This resolves #1062.


jamesaoverton (approved):
Perfect, thanks!

PR #1372, opened 2021-06-15
Update parent & definition of 'pathogen detection assay'

### LINKED ISSUE #1360: 'pathogen detection assay' should be child of 'organism identification assay'
Currently, these two are siblings. But 'pathogen detection assay' is equivalent to 

assay
 and (has_specified_output some ('is about' some pathogen))
 and (achieves_planned_objective some 'detection of pathogenic organisms objective')

so should be under 'organism identification assay'.
(Perhaps the label should also be changed to 'pathogen organism detection assay' for clarity.)

johnwjudkins:
Closes #1360


jamesaoverton (approved):
I will update the assays.owl file and merge this.

PR #1373, opened 2021-06-15
removes "SPLIT" from parent column 

### LINKED ISSUE #1359: Consider removing "SPLIT" from Parent columns
OBI has been pretty consistent about only allowing one asserted parent term, but our ROBOT templates do allow more than one by specifying a "SPLIT" in the ROBOT template string.

I don't remember why we do with with the ROBOT templates. Maybe there were exceptions we wanted to handle?

hectorguzor:
This resolves #1359 by removing "SPLIT" from the parent column in 3 templates. It also removes 'apron' as a parent of 'protective apron' in `devises.tsv`,  and keeps 'personal protective clothing item' as the parent class. 

As discussed on OBI call 2021-06-14, it is recommended that 'apron' be made obsolete.


ddooley (approved):
All Good!

jamesaoverton (approved):

Issue #1376, opened 2021-06-23
TUNEL assay term request


Vineyardintern:
Preferred name: TdT-mediated dUTP nick end labeling assay NCIT:C19815

Synonym: TUNEL labeling ERO:0000731

Textual Definition: A method to detect apoptosis.
Source: NCI TdT-Mediated dUTP Nick End Labeling Assay ([Code C19815](https://thesaurus.cancer.gov/ncitbrowser/pages/concept_details.jsf?dictionary=NCI_Thesaurus&version=21.03e&code=C19815&ns=ncit&type=properties&key=null&b=1&n=0&vse=null))

Parent Class: 'assay'

Submitter: 
Daniel Olson
https://orcid.org/0000-0002-8134-1207

PR #1377, opened 2021-06-23
Update assays.tsv (training)

Vineyardintern:
Update to issue #1376 

Added new row of 'TUNEL assay' to assays.tsv and pushed to origin. 

@beckyjackson


beckyjackson:
Hi Daniel, please take a look at the diff here (click "Load diff"): https://github.com/obi-ontology/obi/pull/1377/files

As we mentioned in the class a couple of weeks ago, Excel does some weird things to quoting in these TSV templates. Therefore, when you saved the TSV, it updated other lines that weren't actually edited.

GitHub diffs are super important for reviewing changes to the ontology, and it makes it much easier for the OBI developers to review your PR if only the lines you actually edited appear in the diff. The quoting issues also cause `robot template` to fail, so we cannot build the `assays.owl` file without fixing the quoting first.

Since we didn't go into detail of how to handle this, **this PR is OK as is and the diff won't impact your assessment**. But, if you'd like to try fixing it, you can try using the workflow here: https://github.com/obi-ontology/obi#editing-templates-in-excel

Note that if you use the workflow above, you'll want to start with a fresh assays template, not the one that already has your changes (although you can just copy and paste your new term into it). Please let me know if you have any questions whatsoever!

Vineyardintern:
Thank you for the heads-up Becky. I will try to fix this and resubmit if I don't run out of time. The makefile has been fighting me every step of the way. lol

I couldn't figure out why I was getting a nultiple targets error for the longest time and I believe it had something to do with tabs vs spaces maybe? I tried a bunch of things and one of them worked apparently. Also for some reason I couldn't use CURL to download the file I needed. It would continually tell me that I had no URL (?) so I switched to wget and now it refuse to place the any data in my file for some reason.

build/do.owl :
wget "http://purl.obolibrary.org/obo/doid.owl" > $@

Instead it downloads the file and saves it as doid.owl and creates an empty do.owl in my build directory?

Daniel


Sent from Outlook<http://aka.ms/weboutlook>

________________________________

Vineyardintern:
Of course the second I send an email I figured out the issue.

build/do.owl :
wget "http://purl.obolibrary.org/obo/doid.owl" -O $@


I needed the -O option. Please ignore the last email. Thanks.


Sent from Outlook<http://aka.ms/weboutlook>

________________________________

Vineyardintern:
Hi Becky,

I looked through the instructions and I must admit I'm not sure how to make them work correctly.As an alternative, would it be possible to refork the repository, make a new pull request, copy the added row from my assays.tsv file and simply append that row to the end of the 'unmolested' file from the new pull request? Thanks.

Daniel


Sent from Outlook<http://aka.ms/weboutlook>

________________________________

beckyjackson:
That’s a perfectly fine solution as well! Is there something specific in the instructions that you find confusing? I want to make sure those docs are as clear as possible, so any feedback is helpful.

Vineyardintern:
Hi Becky,

To be honest I'm not even sure how this works conceptually.  It looks like I need to run the python statement to install the requirements.txt file and then run a make command (I assume the makefile is somehow downloaded with the requirements?). However, the next part is confusing to me. By "edit the sheet", does it mean to edit the obi.xlsx file in some way and if so what edits do I need to make? Then it asks me to do another make command to update the TSVs. So at this point will I be able to edit the TSV files in excel? Or was the editing supposed to happen at the "edit the sheet" stage?

Instead I download the unmolested TSV file and appended my row and then pushed the forked repository back to the master. Hopefully it worked but I would really like to know how the commands (below) you sent me work. Maybe we can briefly go through them at the next class? Thanks!



"First, install the python requirements:

python3 -m pip install -r requirements.txt


Then, make the Excel sheet:

make obi.xlsx


Edit the sheet, save it, and finally, run the following to update the TSV versions of the templates:

make update-tsv"




Sent from Outlook<http://aka.ms/weboutlook>

________________________________

beckyjackson:
Thanks so much for the feedback! I'm going to slightly update the documentation so it's more clear about what you're editing.

>  By "edit the sheet", does it mean to edit the obi.xlsx file in some way and if so what edits do I need to make?

You should edit `obi.xlsx` that you created when you ran `make obi.xlsx`. When you open this Excel file, you'll see that there is a tab for each TSV. In this case, you'd find the tab "assays" and then add your new term to the end of that tab, just like you did with appending it to the TSV.

> Then it asks me to do another make command to update the TSVs. So at this point will I be able to edit the TSV files in excel? Or was the editing supposed to happen at the "edit the sheet" stage?

You're correct that the editing was supposed to happen in the last step, but I realize "edit the sheet" is not clear that we're talking about the `obi.xlsx` Excel spreadsheet with the TSVs as tabs within it. `make update-tsv` takes `obi.xlsx` and converts all the tabs in the Excel spreadsheet back into TSVs. You don't need to do any more editing at this step.

I'd be happy to spend 10 minutes during our review session going over these. They aren't necessarily applicable to *all* ontology development, but it is a cool way to use Makefiles to make editing easier. I just apologize that the documentation wasn't super clear! Please let me know if what I said above makes sense.

Issue #1322, opened 2021-03-30
NTR: terms from VEuPathDB Ontology of potential general interest


cstoeckert:
In a review of VEuPathDB terms, we identified the following as ones that might fit in OBI. OK to add these?
EUPATH:0010001 | antibiotic administration process
EUPATH:0043192 | egg specimen collection process
EUPATH:0043027 | larval specimen collection process
EUPATH:0043060 | pupal specimen collection process
EUPATH:0043054 | free-standing window trap
EUPATH:0000528 | gametocyte detection assay
EUPATH:0000545 | malaria rapid diagnosis test
EUPATH:0000544 | organismal body temperature measurement assay
EUPATH:0000538 | parasite detection assay
EUPATH:0000724 | stool test for microorganisms
EUPATH:0010098 | average respiratory rate
EUPATH:0010078 | mean height
EUPATH:0010087 | mean mid-upper arm circumference
EUPATH:0021083 | blood specimen in vacutainer
EUPATH:0000127 | blood specimen on filter paper
EUPATH:0010079 | median height
EUPATH:0010088 | median mid-upper arm circumference
EUPATH:0011795 | blood test
EUPATH:0000437 | falcivax rapid diagnostic test
EUPATH:0000438 | optimal-IT rapid diagnostic test
EUPATH:0011797 | urine test
EUPATH:0010102 | water container
EUPATH:0043115 | insecticide resistance datum
EUPATH:0043194 | organism identification datum
EUPATH:0000455 | proportion mapped reads
EUPATH:0043003 | BG-lure scent dispenser
EUPATH:0043048 | hand-held sweep net
EUPATH:0043050 | Centers for Disease Control and Prevention gravid trap
EUPATH:0010036 | information on eligibility criterion
EUPATH:0000333 | easting coordinate measurement datum
EUPATH:0000334 | northing coordinate measurement datum
EUPATH:0000433 | parasite species identification by PCR assay
EUPATH:0010047 | study center
EUPATH:0000539 | parasite detection by PCR assay
EUPATH:0000603 | sequencing protocol
EUPATH:0000608 | specimen collection protocol
EUPATH:0000606 | participant identifier
EUPATH:0010095 | axillary temperature measurement datum
EUPATH:0000110 | organismal body temperature measurement datum
EUPATH:0000512 | common name of organism
EUPATH:0000120 | age since birth at time of enrollment
EUPATH:0000113 | age since birth at time of visit
EUPATH:0000327 | entomological data item

[...]

PR #1381, opened 2021-06-24
Device terms from VEuPath

johnwjudkins:
Device terms in #1322.


jamesaoverton:
Discussed on 2021-08-09: James will update `devices.owl` and merge.

PR #1382, opened 2021-06-24
Sequence analysis term from VEuPath

johnwjudkins:
#1322.


jamesaoverton:
It seems wrong to say this is "a data item that is about a sequencing assay". Bjoern suggests that it is about the output of a read-mapping process. Can you please revise this definition?

PR #1387, opened 2021-07-02
Second attempt at assays.tsv update

Vineyardintern:


beckyjackson:
It looks like when you merged the template, some of the git lines were included from the merge conflict. Could you just remove these three additional lines at line 834, 845, and 837? These are totally safe to remove and you don't need to make any other edits. Thanks!

Here's some more details about git conflict markers if you're curious: https://docs.github.com/en/github/collaborating-with-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-using-the-command-line

Issue #1379, opened 2021-06-24
Request to import OGMS 'clinical laboratory test'


johnwjudkins:
VEuPath imports [clinical laboratory test](http://purl.obolibrary.org/obo/OGMS_0000056) as a parent to [blood test](http://purl.obolibrary.org/obo/EUPATH_0011795) and similar terms we are submitting to OBI in #1322. This allows terms like 'blood test' to include not only analyte assays but also the counting of cells in a specimen.

zhengj2007:
@johnwjudkins Can we close it?

johnwjudkins:
Closing. 'blood assay' and similar terms are children of 'assay'. The proposed change is not necessary.

PR #1391, opened 2021-07-15
Assay terms from VEuPath

johnwjudkins:
#1322


johnwjudkins:
This pull request includes external ontology terms to use in axioms for 'parasite detection assay' and 'Plasmodium gametocyte detection assay' - this is causing the check to fail.

jamesaoverton:
I will update the imports to try and make this pass the tests.

jamesaoverton:
I updated various imports and modules, but then I got stuck. Some new rows in `assays.tsv` refer to 'clinical laboratory test', which is not in OBI or this PR. I believe this refers to <http://purl.obolibrary.org/obo/OGMS_0000056> mentioned here in #1379. What would you like to do?

zhengj2007:
Discussed on OBI call 2021-08-09. Proposed to change the parent of 'blood test', 'urine test' to 'assay' and make them as defined class, such as input is 'blood specimen' or evaluate is 'blood'.

jamesaoverton:
@johnwjudkins: I believe that @zhengj2007 was suggesting that 'blood test' and 'urine test' be defined using equivalent class axioms. I tried it locally and it seemed to work well. Would you be OK if I make this change?

johnwjudkins:
@jamesaoverton Yes, those changes are fine, thanks.

jamesaoverton:
Thanks!

Issue #1370, opened 2021-06-11
NTR: laboratory-based population


cstoeckert:
This term is requested for the VEuPathDB project and will be imported into the VEuPathDB ontology once available. This will be used instead of MIRO_30000033 as MIRO is inactive.
Thanks,
Chris

editor preferred term: laboratory-based population
textual definition: A population where the collection of individuals is located and maintained in a laboratory, research, or production facility.
definition source for the textual definition: VEuPathDB
logical definition: subclass of ‘population’

jake-emerson:
Would this fit as a research resource ID (https://scicrunch.org/resources)?

cstoeckert:
@jake-emerson If you mean, are some of these laboratory-based populations available to be shared and could be given a resource ID, then yes.

PR #1392, opened 2021-07-19
lab-based population issue 1370

cstoeckert:


cstoeckert:
Addresses https://github.com/obi-ontology/obi/issues/1370

PR #1394, opened 2021-07-22
Add IDO 'antibacterial'

johnwjudkins:
This is related to the other PRs associated with issue #1322. It was suggested during the last call that 'antibiotic administration process' should include an axiom linking this class to an antibiotic class, so I'm proposing:
`realizes some ('material to be added role' and ('role of' some antibacterial))` ('antibacterial' being an IDO material entity class).

If this PR adding the antibacterial class is accepted, I will add the axiom to the PR I made for obi-edit.
(Edit: wrong role in axiom - does not affect this pull request)


zhengj2007:
I think it should be `realizes some ('material to be added role' and ('role of' some antibacterial))`

jamesaoverton:
Discussed on 2021-08-09: James will update IDO import and merge this.

Issue #1399, opened 2021-08-01
Generalize "cytometric bead array assay" definition and add two subclasses


hectorguzor:
The definition of "cytometric bead array assay" (OBI:0000920) is: 

"An analyte assay in which a series of beads coated with antibodies specific for different analytes and marked with discrete fluorescent labels are used to simultaneously capture and quantitate soluble analytes."

According to https://www.luminexcorp.com/xmap-technology/#applications beads can be coated with molecules other than antibodies. So as it stands the definition seems too narrow.  

**Proposed generalized textual definition**: An analyte assay in which a series of beads are coated with antibodies, antigens, or other small molecules in order to detect an analyte of interest. 

In addition, the CMI-PB project is requesting the following new terms:

**Antigen cytometric bead array assay**
**textual definition**: A cytometric bead array assay in which a series of beads are coated with antibodies specific to antigens of interest. Antigens of interest attach to antibodies on the beads. Fluorescent labeled secondary antibodies are used to detect the antigens of interest. 
**logical definition**:   _subclass of_ = cytometric bead array assay
                                  _target entity_ = antigen

**Antibody cytometric bead array assay**
**textual definition**: A cytometric bead array assay that detects antigen-specific antibodies. In the assay a series of beads are coated with antigens which capture antibodies specific to the antigens. Fluorescent labeled secondary antibodies are used to detect the antibody of interest. 
**logical definition**:   _subclass of_ = cytometric bead array assay
                                _target entity_ =  immunoglobulin complex
                                _output_ =  Antibody titer (import from VO) http://purl.obolibrary.org/obo/VO_0000150

DanBerrios:
Consider slight rewording of labeling:  "Antibody assay by cytometric bead array"  or "Cytometric bead array antibody assay" and similar pattern for antigen assay.  Other assays in the OBI have used the former pattern of <measured thing> by <method>

DanBerrios:
@hectorguzor @rvita If it's a quantitative ab level, consider "Antibody concentration by cytometric bead array assay".  If it's binary, "Antibody detection by cytometric bead array assay".

hectorguzor:
Thanks @DanBerrios. We will label the terms as follow: "antigen detection by cytometric bead array assay' and "antibody detection by cytometric bead array assay".  We will also remove 'Antibody titer' as output of "Antibody detention by cytometric bead array assay" since assay is about detection only.

PR #1401, opened 2021-08-03
cytometric bead array assays

hectorguzor:
This generalizes the definition for 'cytometric bead array assay' and adds two subclasses as discussed in issue #1399. It also imports 'antibody titer' from VO.


jamesaoverton:
We will run this past @rvita

jamesaoverton:
We're going to make a fresh PR for this.

Issue #1397, opened 2021-07-28
Request to import terms from COB


hectorguzor:
The IEDB is requesting to import two terms from COB. We are requesting to import 'molecular entity' (COB:0000013) and 'complex of molecular entities' (COB:0000080). These terms are needed to create the terms for issue #1380. Thanks.

Issue #1380, opened 2021-06-24
NTR: 3D determination assays and material entity terms


hectorguzor:
These terms are being requested for the work the IEDB is doing with some immune epitope assays. This is related to issue #1346. As of now, we don't have a good place for the material entity terms being requested, so we are asking to place them in OBI temporarily. 

Thanks

![image](https://user-images.githubusercontent.com/78870607/123308807-fe55ea80-d4d8-11eb-8856-d5b37f151e22.png)

hectorguzor:
Sorry, I forgot to also add the following term to the list:

5) **MHC:Ligand complex**
     **Textual Definition**: A complex of molecular entities which is the result of a MHC molecule binding with a ligand. 
     **Logical Definition**: (SubclassOf 'complex of molecular entities') and ('has part' some 'MHC 
                                    protein complex')

PR #1402, opened 2021-08-04
Imports COB term, adds 3 molecular complex terms, and 2 assays

hectorguzor:
This imports 'complex of molecular entities' from COB as requested in #1397, and creates 5 new terms as discussed in #1380.


jamesaoverton:
Hi @hectorguzor. The diff for `obi-edit.owl` is bigger than it should be. You can see that trailing whitespace is being trimmed from a bunch of annotations that have nothing to do with your changes. I think I do actually want to keep those changes. But keep an eye out for that sort of thing going forward because it usually indicates a problem.

jamesaoverton:
Discussed on 2021-08-09: This is just importing one COB term 'complex of molecular entities', which we need to improve the definitions of some immunology assays. This is a term that doesn't really exist elsewhere in OBO, and is not a good fit for ontologies other than COB. For example, ChEBI has 'chemical substance' which is too general. Protein Ontology has complexes of proteins, which is too specific. I believe that GO doesn't want this term either. But it is important for immunology and thus the large number of immune assays in OBI. So of all the COB terms, I think this one is safe to import.

jamesaoverton:
There ended up being a few problems with this PR. I want to drop it and try again. @hectorguzor could you please make three new PRs, one-by-one, and I'll try to merge them quickly: (1) the COB import, (2) the 3 molecular complex terms, and (3) the 2 new assays. Sorry for changing my mind about this!

Issue #1123, opened 2020-03-09
Duplicate definitions for inverse properties


cstoeckert:
Discussed on March 9, 2020 OBI call. Duplicate definitions were identified in the ROBOT report for:
has_specified_output [OBI:0000299] and is_specified_output_of [OBI:0000312]
has_specified_input [OBI:0000293] and is_specified_input_of [OBI:0000295]
Instead of repeating the definition, should just define as the inverse property as in:
has_specified_input is the inverse property of is_specified_input_of

ddooley:
Done in pull request https://github.com/obi-ontology/obi/pull/1406

PR #1406, opened 2021-08-16
duplicate specified i/o definition fix

ddooley:
Issue #1123 has_specified_input and has_specified_output duplicate definition of inverse relation.  Tweaking this.


jamesaoverton:
I think these two terms are also coming from the IAO import, so we get both the old and the new proposed definitions, which fails our test for multiple textual definitions. The solution would be to (1) edit the IAO import, as a temporary fix, and (2) update IAO, as a permanent fix.

jamesaoverton:
Discussed on the 2021-08-16 call: No objections. This is a little tricky. I'll work on it when I get a chance.

jamesaoverton:
I'm tempted to just use a SPARQL query to remove the old definitions somewhere late in our build process. @beckyjackson can you please give that a try?

beckyjackson:
@jamesaoverton I added a SPARQL update to remove the bad definitions from the merged version of OBI. This fix will only appear in the OBI release, not in the edit file.

Issue #1364, opened 2021-06-02
NTRs for 3D neuron morphology determination assays


patrick-lloyd-ray:
Following discussion with @hectorguzor, I'd like to submit the following NTRs: https://docs.google.com/spreadsheets/d/1Nc6rZwuUvNtrv4GOQBMckNmXggWhM9_lHJKD1r8xe_E/edit?usp=sharing

Any comments, suggestions, revisions, and the like are welcome.

hectorguzor:
Discussed in OBI call 2021-08-16: Consider putting neuron assays under 'cytometry assay' as inferred classes. Patrick and Hector will discuss this as they work on the PR.

PR #1410, opened 2021-08-17
NTR 3D neuron morphology assays issue 1364

hectorguzor:
This imports 'neural cell' term from Cell Ontology, and creates 5 new 3D structure determination assays to close issue #1364.


mbrochhausen (dismissed):
Per OBI call on Aug 30: Looks good, but the textual definitions should reflect the asserted parent.

PR #1421, opened 2021-08-31
Immune epitope assays

hectorguzor:
This modifies "Immune epitope assay" and its direct subclasses.


jamesaoverton:
Discussed on the OBI call 2021-09-20: This only affects the 'immune epitope assay' branch, and mostly updates the logic while keeping the structure largely the same. Nobody has expressed any concerns, so I intend to merge this.

PR #1424, opened 2021-09-06
80 assays for CEBS

jamesaoverton:
This PR creates 80 new clinical assays for the CEBS project, and required imports. We have coordinated with TURBO on the modelling, e.g. in #1327.

One blocking problem is that I don't know where to put these to Human Phenotype terms:

- http://purl.obolibrary.org/obo/HP_0020082 Heinz bodies
- http://purl.obolibrary.org/obo/HP_0032550 Howell-Jolly bodies

In HP they're descendants of 'Phenotypic abnormality', which I don't know how to place in OBI.


jamesaoverton:
I'm chairing the next OBI call, and I'd like to discuss these terms then. Feedback from the TURBO team (@cstoeckert and maybe @johnwjudkins) would be appreciated.

cstoeckert:
FWIW we placed phenotypic abnormality under quality in the VEuPathDB ontology and TURBO ontology.  Although they have a material basis, Heinz bodies and Howell_Jolly bodies do have a specific dependency  as red blood cell inclusions.

jamesaoverton:
Discussed on OBI developer call 2021-09-13:

- @rvita will make introduction with Clinical measurement ontology (CMO) people from the Rat Genome Database (RGD); but we will continue putting these in OBI
- in the textual definitions for all these assays we should use "specimen" (more general) rather than "sample" (OBI is specific about the meaning of "sample")
- we need to add editor information
- for percentage assays, it would be better to have "percentage value specification" as the output; @cstoeckert suggests this but says it is not a blocker; we could resist this later
- we'll consider generalizing 'reticulocyte count per 1000 red blood cells assay' to just be a concentration
- 'CBC with manual differential venous blood myelocyte count assay' would be better as a child of the new 'myelocyte count assay'; we had some discussion of CBC and their parthood

Regarding the HPO terms, we would like to solve the problem very generally at the COB level. See https://github.com/OBOFoundry/COB/issues/123

We could change our modelling of these terms to be specific: Counting cells with the HPO phenotype. Then we put the two HPO terms under 'quality'.

rvita:
I emailed Stanley Laulederkind and Shur-Jen Wang of RGD and cc'd Jennifer.

jamesaoverton:
I dropped these three terms from the PR:

- OBI:0003169 reticulocyte count per 1000 red blood cells assay
- OBI:0003178 band neutrophil percentage assay
- OBI:0003179 band neutrophil count assay

Other changes:

- in the textual definitions "sample" has been changed to "specimen"
- term edit has been added

Still to do:

- [] make 'CBC with manual differential venous blood myelocyte count assay' a child of the new 'myelocyte count assay'

I will drop these four from the current PR and not include those HPO terms:

- Heinz bodies assay
- Heinz bodies percentage assay
- Howell-Jolly bodies assay
- Howell-Jolly bodies percentage assay

jamesaoverton:
Discussed on the OBI call 2021-11-08: James will address these points and then merge.

PR #1433, opened 2021-09-23
RNA sequencing subclasses

hectorguzor:
Makes 'direct RNA sequencing assay' and 'PRO-cap'  children of 'RNA sequencing assay' as discussed in issue #1395.


jamesaoverton:
Discussed on OBI call 2021-09-27: This PR make a small change that we do want to make, even though it doesn't fully address #1395.

Issue #1416, opened 2021-08-25
NTR for IEDB: binding datum and assays 


hectorguzor:
The IEDB team is requesting the following terms to better organize 'binding datum':

**categorical binding datum**
_textual definition_: A binding datum about two or more material entities to form a complex that is expressed as a categorical value. 
_editor preferred term_: qualitative binding datum
_logical definition_:  subclass of = binding datum and 'has value specification' some 'categorical value specification' 

**quantitative binding datum**
_textual definition_: A binding datum about two or more material entities to form a complex that is expressed using numbers.
_logical definition_: subclass of = binding datum 

we are also requesting the following assays:

**assay measuring quantitative binding of a MHC:ligand complex**
_textual definition_: An assay that detects the binding of a MHC molecule with a ligand, and produces a quantitative measurement of the binding as an output. 
_logical definition_: subclass of = assay 
_target entity_ = 'MHC protein complex binding to ligand' 
_output_ = 'quantitative binding datum' 

**assay measuring quantitative binding of a antigen:antibody complex**
_textual definition_: An assay that detects the binding of an antigen with an antibody and produces a quantitative measurement of the binding as an output. 
_logical definition_: subclass of = assay
_target entity_ = 'immunoglobulin binding to epitope'
_output_ = 'quantitative binding datum'

**assay measuring quantitative binding of a epitope:MHC:TCR complex** 
_textual definition_: An assay that detects the binding of a MHC:epitope complex with a T cell receptor, and produces a quantitative measurement of the binding as an output.
_logical definition_: subclass = assay 
_target entity_ = 'MHC:epitope complex binding to TCR'
_output_ = 'quantitative binding datum'

hectorguzor:
Also requesting the following two quantitative binding datum terms:

**EC50 binding datum**
_textual definition_: A quantitative binding datum that is expressed as a half-maximal effective concentration (EC50).
_logical definition_:   equivalent to = quantitative binding datum and ((half maximal effective concentration (EC50)) and ('is about' some binding))

**IC50 binding datum**
_textual definition_: A quantitative binding datum that is expressed as a half-maximal inhibitory concentration (IC50).
_logical definition_: equivalent to = quantitative binding datum and ((half maximal inhibitory concentration (IC50)) and ('is about' some binding))

bpeters42:
A few edits in the below:

categorical binding datum
textual definition: A binding datum that is expressed as a categorical value.
editor preferred term: qualitative binding datum
[...]

PR #1434, opened 2021-09-24
Binding datum and assays

hectorguzor:
This PR adds 4 binding datum terms and 3 quantitative binding assays to close issue #1416.


jamesaoverton:
Discussed on the OBI call 2021-09-27. I added notes to the issue.

jamesaoverton:
Discussed on the OBI call 2021-10-04: There are still harmonization problems but they are bigger than this one issue, and we should not get stuck. So I plan to merge this.

PR #1435, opened 2021-09-24
Replaces hospital with 'hospital organization' from OMRSE.

### LINKED ISSUE #1295: Discussion of term Hospital and similar terms in OMRSE
See https://github.com/ufbmi/OMRSE/issues/153

johnwjudkins:
Closes #1295. This pull request also adds 'hospital role' since it is referenced in the definition of 'hospital organization'.


jamesaoverton:
- We stopped using underscores in labels years ago, so I would prefer 'obsolete hospital' as the label.
- We should include an obsolescence reason. I suggest [IAO:0000228 'term imported'](http://purl.obolibrary.org/obo/IAO_0000228).
- We should use the 'replacement term' column to specify OMRSE:00000056 (hospital organization)
- We should preserve the term editors.

PR #1449, opened 2021-10-20
Revision for new vial hierarchy issue #1229

### LINKED ISSUE #1229: vial and chromatography device
Vial has parent under chromatography device at moment but should have primary parent "container". 

"Chromatography device" should be under device, but it is currently under physical store.

ddooley:
Fixes #1229

Requires NTR "vial" , renaming existing "vial" -> "chromatography vial", and relocation of "test tube"


jamesaoverton:
@ddooley When you have time, please address my comment here https://github.com/obi-ontology/obi/issues/1229#issuecomment-963205449

PR #1468, opened 2021-12-13
Remove whitespace around pipes in templates

jamesaoverton:
- antibodies from lymphocyte supernatant assay
- Remove whitespace around pipes in templates


jamesaoverton:
@beckyjackson Would it be reasonable to add a check for this? What about making it part of the sort script?

Issue #1393, opened 2021-07-19
CFDE NTRs for HuBMAP project


mgiglio99:
Hello,
We are working on Batch 2 of term requests as part of the CFDE project. This set is needed by the HuBMAP project.
Here is a list of the terms with draft definitions, parentage, etc.
https://docs.google.com/spreadsheets/d/1_XaxKYQKO0nnH8yY8zwz_lhkweuUsO0ARyF22PI1uTk/edit?usp=sharing
We are working on finalizing them, but I feel the terms are far enough along to reserve OBI term ids.
Please take a look and see if you agree that they are in scope for OBI and that it would be ok to register OBI ids for them at this stage.
Thanks,
Michelle and @nsuvarnaiari

hectorguzor:
Hi @mgiglio99,

here is the definition for 'detection technique' I showed in the OBI call on  7/26/21.

Detection technique =(def) Assay **D** counts as a detection technique in assay **A** if and only if 

(I) In generating information _ia_ about target _ta_  of **A**, one utilizes **D** to generate information _id_ about a proxy _td_, where _id_ is the target of **D** and _id_ is used to determine what a measurement of _ia_ would be. And 

(II) _td_ and _ta_ are disjoint. 

in looking at the terms you requested:

1. what you have for row 10 as a detection technique seems fine to me. It is the fluorescence detection assay that detects the fluorescently labeled molecules to get at localizing specific DNA or RNA sequences. 
2. Row 7 doesn't have a textual definition, but I would imagine (given the label) that identifying an antibody for imaging will be done in virtue of detecting fluorescence. 
3. Row 4 is an interesting case. Because the parent for "autofluorescence microscopy assay" is "autofluorescence microscopy assay" (which has as a parent "fluorescence detection assay"). My intuition is to leave "detection technique" empty here. Because while a microscope is used to produce a magnified image of the object, I'm assuming the object is still detected via a fluorescence detection assay. (but I could be wrong).
4. for row 18 I am also not sure. This may be fine. sometimes it seems to me that mass spectrometry assays are detection techniques and other times not. I'm not sure what to think in this case. It seems to work here. 

I hope this helps

mgiglio99:
Thanks for this Hector - I'll take a look at these more closely - this is great input as we are still working on developing these.
Another set that we've already done the pull request for and that we feel are ready to and that we'd like to get your input on are in https://github.com/obi-ontology/obi/issues/1243
Could you take a look at those ?
Thanks,
[...]

PR #1471, opened 2021-12-17
New terms & updates for CFDE

beckyjackson:
This PR contains new terms from https://github.com/obi-ontology/obi/issues/1393 as well as some label updates - see Google sheet here: https://docs.google.com/spreadsheets/d/1_XaxKYQKO0nnH8yY8zwz_lhkweuUsO0ARyF22PI1uTk/edit#gid=0

Some notes from @nsuvarnaiari on the Google sheet above concerning the changes in this PR:
1. Most of the terms are new but there are some old terms with new parentage/new definitions/new synonym. The new terms have provisional ids which are given in ColA.
2. Row 3: fluorescence microscopy assay : Is an old term for which we are adding subclass axiom instead of equivalence axiom. More details on it in the ticket (from Hector on Nov 22nd).
3. Row 33: chromosome organization assay by fluorescence in-situ hybridization:  I have changed the parent of the term from “in-situ hybridization assay” to “fluorescence in-situ hybridization assay” which is a new term in the template at row 11. We want to remove the “alternate term” for this term - “FISH”  as it is the alternate term for it’s parent- “fluorescence in-situ hybridization assay”.
4. Rows 29-32 are just the label changes. Added suffix “assay” to the terms.

On row 27, "single-nucleus chromatin accessibility and mRNA expression sequencing assay" - we would like to have this term be a child of both:
* single-nucleus RNA sequencing assay (current asserted parent)
* single-nucleus ATAC-seq

We would need to either add an equivalence axiom to one of these terms, or assert two parents (there is no SPLIT in the assay template right now, and we typically don't assert two parents). I believe @mgiglio99 is OK with this single parent for now until we figure out how best to proceed.


src/ontology/templates/assays.tsv
```diff
@@ -240,6 +240,7 @@ OBI:0002112	small RNA sequencing assay		http://purl.obolibrary.org/obo/IAO_00001
...
 OBI:0002115	cytometry time of flight assay			CyTOF|mass cytometry assay	A cytometry assay in which the presence of molecules of interest on or in cells is indicated by binding of antibodies labeled with rare earth element tags which are detected by time-of-flight mass spectrometry.	PMID:26190063	Whole human blood samples were incubated with lipopolysaccharide (LPS) to examine dose-dependent signaling responses within the toll-like receptor 4 (TLR4) pathway. After LPS stimulation, rare earth metal tagged antibodies against phenotypic markers were used to stain the blood sample (Nd143-CD45RA, Nd145-CD4, Er170-CD3, Sm152-TCR__, Pr141-CD7, Nd146-CD8, Nd142-CD19). A Fluidigm CyTOF version 1 instrument equipped with CyTOF software version 5.1.648 was used to measure and analyze the levels of antibody staining refelctive of the level of expression of those surface receptors.	ImmPort					subclass	logical definition does not capture ""time of flight"" part	cytometry assay		mass spectrometry assay						rare earth metal atom					
+OBI:0002115				CyTOF|mass cytometry assay				Michelle Giglio|Suvarna Nadendla|Rebecca Jackson			CFDE	https://github.com/obi-ontology/obi/issues/1402	subclass
```
jamesaoverton, line 243: Something is wrong here. There are two rows with ID "OBI:0002115", and the second is missing a label. We should have a test to catch this.

nsuvarnaiari:
@jamesaoverton 

The second row for OBI:0002115 that is without a label was only to add an alternative name. Otherwise both the lines - 242 and 243 should be the same.

@mgiglio99

Thanks,
Suvvi

PR #1478, opened 2022-01-10
Never allow equivalent classes

### LINKED ISSUE #1477: Automatically reject equivalent named classes
This was raised by @cmungall in response to PR #1474, which led to issues #1475 and #1476. For OBI, I think that if the reasoner finds that two named classes are equivalent, this is never intentional and should always be reported as an error.

ROBOT can easily catch these problems: <http://robot.obolibrary.org/reason#equivalent-class-axioms>. I think we should have been using this ROBOT features since it was available, and failing to use it was just an oversight on my part.

So I suggest using ROBOT's `--equivalent-classes-allowed never` in both our ELK test and HermiT release tasks. Including it in our ELK tests will mean that PRs that introduce bad equivalents will fail the automated tests, which would have caught #1475 and #1476.

beckyjackson:
Resolves #1477, but this will fail until #1475 and #1476 are resolved.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,6 +181,7 @@ obi.owl: build/obi_merged.owl
 	$(ROBOT) reason \
 	--input $< \
 	--reasoner HermiT \
+	--equivalent-classes-allowed none \
 	annotate \
 	--ontology-iri "$(OBO)/obi.owl" \
 	--version-iri "$(OBO)/obi/$(TODAY)/obi.owl" \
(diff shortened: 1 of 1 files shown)


ddooley:
We will tweak biobank template in https://github.com/obi-ontology/obi/issues/1476 to fix this.

jamesaoverton:
With #1661 we can now fix this. I'm going to create a new PR.

Issue #1482, opened 2022-01-11
NTR: cytotoxic T cell degranulation terms 


hectorguzor:
The IEDB is requesting 4 terms in order to remodel assays under [assay measuring epitope specific cytotoxic T cell degranulation](http://purl.obolibrary.org/obo/OBI_0001269). These will be subclasses of [cytotoxic T cell degranulation](http://purl.obolibrary.org/obo/GO_0043316).

- **perforin release by T cell**: A cytotoxic T cell degranulation that consists of the release of perforin.
- **granzyme A release by T cell**: A cytotoxic T cell degranulation that consists of the release of granzyme A.
- **granzyme B release by T cell**: A cytotoxic T cell degranulation that consists of the release of granzyme B.
- **granulysin release by T cell**: A cytotoxic T cell degranulation that consists of the release of granulysin.

addiehl:
Sorry for veering out of my lane as well as not seeing this before, but I really don't like the wording of your definitions, which read as if there are four types of cytotoxic T cell degranulation, each of which releases just one kind of mediator.  In my view, both the definition and the parent class are wrong.

I would recommend the grandparent class to be [production of molecular mediator of immune response (GO:0002440)](http://amigo.geneontology.org/amigo/term/GO:0002440), defined as "The synthesis or release of any molecular mediator of the immune response, resulting in an increase in its intracellular or extracellular levels"

Under GO:0002440 we have the following 
[perforin production (GO:0035944)](http://amigo.geneontology.org/amigo/term/GO:0035944)
[granzyme A production (GO:0035746)](http://amigo.geneontology.org/amigo/term/GO:0035746)
[granzyme B production (GO:0071613)](http://amigo.geneontology.org/amigo/term/GO:0071613)
[granulysin production (GO:0036262)](http://amigo.geneontology.org/amigo/term/GO:0036262)

Then the definition of perforin release by T cell would be:
"The release of perforin during cytotoxic T cell degranulation"
with the logical definition
'perforin production' and 'part of' some 'cytotoxic T cell degranulation'

The other three terms should be modeled in a similar fashion.
-- Alex

hectorguzor:
Discussed 2022-02-14:

There was agreement on the call about the modeling suggested by Alex. However, this may not work for the modeling of cytotoxic T cell degranulation assays the IEDB needs. The IEDB team will look into this and discuss it further. 

[...]

PR #1496, opened 2022-02-10
adds 4 T cell degranulation terms

hectorguzor:
This PR adds 4 cytotoxic T cell degranulation terms to close issue #1482.


addiehl:
See my comment to #1482 for a different way of modeling these terms.

hectorguzor:
the last two commits make changes according to @addiehl suggestions in issue #1482.

PR #1521, opened 2022-03-26
Detection Technique

### LINKED ISSUE #1488: Detection Technique Definition
Discussed during OBI 2022 workshop:

We need a clear definition of what counts as a detection technique. We decided on the following tentative definition. 

assay B is a _detection technique_ in assay A if:

1. assay B is a part of assay A. and 
2. the target entity of assay B and the target entity of assay A are disjoint. and 
3. if either:
* a. the data item output of assay B **is_proxy_for** the data item output of assay A. or 
* b. the data item output of assay A is the specified output of some [data transformation](http://purl.obolibrary.org/obo/OBI_0200000), with the output of assay B as a specified input.

**What we need to do moving forward:**

1. clarify whether 'detection technique' is a class or a design pattern/ annotation property/ 'role'.
2. clarify how is_proxy_for is being used in the definition. this could be interpreted as a 'no data transformation re-interpretation' of a data item, based on knowing how it was constructed. do we want to use OBI [is_proxy_for](http://purl.obolibrary.org/obo/OBI_0000314)? or something else?
3. figure out how to specify the  'data transformation' part for assays that have one.
4. figure out how to implement logical definition into assays template to identify incorrect techniques. and how to document detection technique definition for future development if the implementation of logical definition into template is not possible.

hectorguzor:
This PR closes #1488 
It adds a 'detection technique' assay class to `assays.tsv`. No assay children will be asserted or inferred under this new class.
The PR also makes changes to the logical axiom of the detection technique column in `assays.tsv` and `epitope-assays.tsv`.


jamesaoverton:
I checked this an it looks good to me. Do we want more feedback before merging?

hectorguzor:
No, I think it's ready to merge. Thanks.

PR #1527, opened 2022-04-14
Created data-sets template and added image data sets from MRIO

adbartni:
Using templates where applicable


jamesaoverton:
Thanks! I will take a look soon.

cstoeckert:
@adbartni we can discuss this on the June 6, 2022 OBI dev call. Can you attend?

adbartni:
@cstoeckert That works for me, I'll be there. I'll discuss with Alex Diehl and others to see if they can join as well.

cstoeckert (commented):
These comments are based on checked out files in image-data-set-template-1481 branch.
There appears to be a minor change in 'magnetic resonance imaging assay' that might be a typo: "|" added to end of term editor. 
Also for 'magnetic resonance imaging assay':
- Values of input, output, and associated axiom are not showing up in obi-edit import (do show up in assays. owl)
- What's the relation to OBI:0000623 ' NMR spectroscopy assay?

data-sets template: 
- definitions don't match asserted parent class in many cases (use more general class)
- definition source column is empty.

obi-edit does not have data-sets owl as an import. But some of these classes are already in obi-edit. (OBI:0003327 - OBI:0003334). 
should add mention of new data-sets template to README.md

adbartni:
So sorry for the delay, I've had to juggle a few projects for my thesis lately. I'll jump on the OBI developer call today to discuss, since it looks like there's time allotted for discussing open PRs on the agenda.

adbartni:
I've managed to sort out the conflicts with the up to date obi-edit.owl and assays.owl/tsv files and am trying to keep the changes confined to templates as much as possible this time around.

Since the edit I want to make to `magnetic resonance imaging assay` is to add an axiom stating that its specified output is a `magnetic resonance image dataset`, the tests will fail because the assays template is processed first and the `magnetic image dataset` class is never inserted. I'd gotten around this the first time I submitted this PR by added a blank class to obi-edit.owl for `magnetic resonance image dataset` to allow the test to run, but as I mentioned I'm trying to keep the changes as much as possible in the templates.

My question before proceeding is if these changes so far can be merged so that `magnetic resonance image data` at least is in obi-edit.owl so that I can properly add the specified output axiom to `magnetic resonance imaging assay`. Or should I just do what I did before and add the bare class for `magnetic resonance image dataset` directly to obi-edit.owl and let the template update the axioms?

bpeters42:
Discussed on 8/29 call: Agreement to add the needed classes to obi-edit.owl first. Minor edits were suggested to call it 'imaging data set', and to add examples (including what is not covered).

adbartni:
@jamesaoverton We've added the examples of usage to our new terms. Before merging, we'd like to discuss the renaming of 'image data set' to 'imaging data set' and our concerns on the 11/21 developer call if there's time.

adbartni:
@bpeters42 We discussed on the obi-developer call today, but wanted to get your opinion before moving forward.

We would prefer keeping the label 'image data set' as opposed to 'imaging data set.' Although some image data sets are the direct result of an imaging assay, within MRIO we have subclasses of 'image data set' (such as 'computed image data set,' 'image segmentation map') that are derived from a 'data transformation' of an existing 'image data set.' As such, we think it makes more sense to leave the label 'image data set' so that there is no confusion whether 'image data sets' are only produced by an imaging assay.

bpeters42:
My concern was that the definition of 'image data set' was not general and didn't seem to match the label. If the definition is broad in line with the label, I have no concerns.

adbartni:
We have revised the definition of 'image data set' to better convey what our class is intended to represent in reality.

Original definition: "A data set that is comprised of structured measurements about some entity and its associated metadata using pixels (2D), voxels (3D), or an arbitrary number of dimensions. An image data set can be the source from which an image is produced."

New definition: "A data set that is comprised of multidimensional structured measurements and metadata required for a morphological representation of an entity. An image data set can be the source from which an image (such as a 2D image using pixels or a 3D image using voxels) is produced."

johnwjudkins:
2023-02-06 @bpeters42 Please review (definition changed).

jamesaoverton:
I'm very sorry that this has taken so long. We will get this PR merged as soon as possible.

One more thing: My understanding is that you prefer the label 'image data set', but the PR still uses 'imaging data set' throughout. Should all 'imaging data set' strings in the PR be changed to 'imaging data set', which would also mean changing 'magnetic resonance imaging data set' to 'magnetic resonance image data set' and 'raw imaging data set' to 'raw image data set'?

src/ontology/templates/assays.tsv
```diff
@@ -417,7 +417,7 @@ OBI:0002968	positive/negative ion switching metabolite profiling assay			positiv
...
-OBI:0002985	magnetic resonance imaging assay			MRI	An imaging assay in which nuclear magnetic resonance is used to produce information about the interior structure and composition of an input material entity.	url:https://en.wikipedia.org/wiki/Magnetic_resonance_imaging		Rebecca Jackson				https://github.com/obi-ontology/obi/issues/1290	subclass		imaging assay						NMR instrument								
+OBI:0002985	magnetic resonance imaging assay			MRI	An imaging assay in which nuclear magnetic resonance is used to produce information about the interior structure and composition of an input material entity.	url:https://en.wikipedia.org/wiki/Magnetic_resonance_imaging		Rebecca Jackson|Alexander D. Bartnik, ORCID:0000-0001-9676-7377|Lucas M. Serra, ORCID:0000-0002-2104-0568|Alexander D. Diehl, ORCID:0000-0001-9990-8331|Mackenzie T. Smith, ORCID:0000-0002-9821-4132|William D. Duncac, ORCID:0000-0001-9625-1899|Lauren M. Wishnie, ORCDI:0000-0002-7245-3450				https://github.com/obi-ontology/obi/issues/1290	subclass		imaging assay						NMR instrument			 ('has role' some 'magnetic resonance imaging evaluant role')					(realizes some (concretizes some 'magnetic resonance pulse sequence protocol'))
```
cmungall, line 420: - typo "Duncac"
- typo "ORCDI"

adbartni: Thanks for catching! I corrected them in the latest commit.

adbartni:
@jamesaoverton I can't make the meeting today so I figured I'd ping you here.

I updated the labels to 'image data set' from 'imaging data set' for the three terms in the data-sets template. Let me know if there's anything else that should be done, but I think it should be ready for merging.

PR #1531, opened 2022-04-20
fluorescence & radioactivity epitope binding assays

hectorguzor:
This PR changes the parent of all radioactivity epitope binding assays from 'radioactivity detection' to 'binding assay using radioactivity detection.' And it changes the parent of all fluorescence epitope binding assays from 'fluorescence detection assay' to 'binding assay using fluorescence detection.'


jamesaoverton:
@hectorguzor please resolve these conflicts

Issue #1227, opened 2020-08-31
remote patient assessment


miseminger:
I'd like to submit "remote patient assessment" or "telehealth patient assessment" as a subclass of “performing a clinical assessment” http://purl.obolibrary.org/obo/OBI_0600003.  This term would refer to a clinical assessment performed remotely by technological means (telephone or video).  Thanks!

ddooley:
label: remote patient assessment
parent: performing a clinical assessment
definition: A clinical assessment performed remotely by technological means
definition source: 
example of usage: A telephone or video remote patient assessment conducted between a physician and patient.

I'll get Madeline to lookup def source.

zhengj2007:
@ddooley FYI: guidelines for OBI term annotations, http://obi-ontology.org/page/OBI_term_guidelines

miseminger:
definition source: Calton, B., Abedini, N., & Fratkin, M. (2020). Telemedicine in the Time of Coronavirus. Journal of pain and symptom management, 60(1), e12–e14. https://doi.org/10.1016/j.jpainsymman.2020.03.019  

term editor: Madeline Iseminger (https://orcid.org/0000-0002-0548-891X)

ddooley:
OBI May 2 call: About interviewee who has assistance?

New parent: organism diagnostic process
Revised definition: "An organism diagnostic process which involves a remote assessment of a patient by a clinician."

ddooley:
This has been completed!

PR #1534, opened 2022-05-02
added remote patient assessment

ddooley:
An NTR for https://github.com/obi-ontology/obi/issues/1227


ddooley:
I think this is ready to go.

PR #1542, opened 2022-05-18
Require passing report on obi-base

jamesaoverton:
This PR adds a task to the Makefile that generates a ROBOT report on the `views/obi-base.owl` file.﻿ This task will exit with a failure code if the report includes any ERROR level messages, and it will print the first ten problems found to the console. The `make test` task now requires this report, so our GitHub Actions continuous integration tests will fail if a PR introduces new ROBOT report ERRORs.


jamesaoverton:
This should pass once the base file is updated to reflect fixes on `master`.

jamesaoverton:
Discussed on the OBI development call 2022-05-23: No objections. This won't catch everything, but it is still progress.

PR #1550, opened 2022-06-02
target context column

### LINKED ISSUE #1516: Add 'target environment' column to assays template
The IEDB team is trying to figure out a way to differentiate between in vivo and in vitro assays.
 
In making this distinction what we are really trying to get at are the environments in which measurements are made, whether it is in a live mouse, test tube, etc. 
 
One possible way to capture this in how we model assays is to specify the target environment in which a target entity is being evaluated. Perhaps we can add a new column to the assays template where we can specify these environments. We can call the column  **target environment** and defined it as:
 
the material entity in which the target entity of an assay is measured, such as a live mouse,  test tube, etc.

hectorguzor:
This PR closes #1516.
It creates 'has assay target context' relation and adds 'target context' columns to `assays.tsv` and `epitope-assays.tsv`.


cstoeckert (commented):
How about specifiying domain = assay and range = material entity?

hectorguzor:
Hi @cstoeckert 
for the new relation I specified range = material entity in `obi-edit.owl`, but I left the domain unspecified.

The axiom `C has_specified_output some ('is about' some ( 'has assay target context' some %))` is meant to indicate that the target entity  'has assay target context' some material entity. But a target entity can be many things. A material entity, a process, a quality, etc. This is why I left the domain unspecified.

hectorguzor:
This conflict is too big. I'm afraid the changes I need to make to the templates could break things in the owl files. It brings in new assays that have recently been merged, and these assays don't have the tab spaces required for the new column. I will create a new PR that includes the new assays and link the comments of this PR to it.

Issue #1546, opened 2022-05-20
NTRs: additional device terms needed for IGS gEAR project


mgiglio99:
Hello,
As part of the the gEAR project at the Institute for Genome Sciences (IGS) we are requesting some device terms for DNA sequencers and arrays.
Here is the link to our draft terms in the device template:
https://docs.google.com/spreadsheets/d/15iyboG5i-Il2w4sfTDjGgtqHn0-Um6VcD8cE66VmFsI/edit?usp=sharing

Thanks,
Michelle, Suvvi, and Jain

cmungall:
Looks like this was done, I see all terms in OBI now - close this (I don't have perms)?

PR #1565, opened 2022-06-29
adding 19 new device terms for gEAR project

nsuvarnaiari:
In this PR, 19 device terms have been added for gEAR project. The ticket referring to this - #1546 

Thanks,
Suvvi ans @mgiglio99


jamesaoverton:
Discussed on the OBI call 2022-07-25: It would be better to remove the "TM" symbols -- Excel (for example) doesn't like them. Suuvi will make this change. Other than that, it looks good!

bpeters42:
Reviewed on 8/29 call; James will merge.

Issue #1556, opened 2022-06-13
Review of Assay with Detection Techniques 


hectorguzor:
We recently  specified the definition of a detection technique as an assay (A) that is part of another assay (B), where the target entities of A and B are disjoint, and the target entity of A is a proxy for the target of assay B.

James created a SPARQL query that returns a TSV file with assays that have a  detection technique. The TSV file has four columns. Column A includes assays that have a detection technique, column B includes these assays’ target entities, column C includes their  detection technique and column D includes the detection technique’s target entities. 

I collected all assays that have target entities that are not disjoint with the target entities of their detection techniques, or where it was not clear to me that there was disjointness because the target entities were not specified. You can fine those assays [Here](https://docs.google.com/spreadsheets/d/1Y_HUi3ZbwDMi2q_HU3stwv7BHh6UxbStLgB66-zFgkc/edit#gid=0). In light of changes to the detection technique columns in the assays template, it would be good to review these assays to ensure the modeling is correct. I started by reviewing 3 assays. These are highlighted in yellow. Column E includes some notes on what I take to be the issue with the modeling  and column F has  notes on how we can go about correcting the modeling. 

For example, in row 4 [FAIRE-seq](http://purl.obolibrary.org/obo/OBI_0001859) has [DNA sequencing assay](http://purl.obolibrary.org/obo/OBI_0000626) as a detection technique. Currently, 'DNA sequencing assay' and 'FAIRE-seq assay' have [region](http://purl.obolibrary.org/obo/SO_0000001) as their target entity. This violates the disjointness condition. In this case I recommend we import  [regulatory_region](http://purl.obolibrary.org/obo/SO_0005836) from SO and make it the target entity of 'FAIRE-seq'. We should also make [primary structure of DNA macromolecule](http://purl.obolibrary.org/obo/OBI_0000811) the target entity of 'DNA sequencing assay'. If we import 'regulatory_region' and make it a sibling of 'primary structure of DNA macromolecule' then we can explain disjointness.

DanBerrios:
Discussed on OBI Dev Call 06/13/22:

- “Regulatory region” should be imported from SO
- “Region” conflates nucleotide and AA sequencing
- Some of the assays on the spreadsheet should have “primary structure of DNA” as target for sequencing assay
- Probably we need to do some remodeling of “primary structure of DNA”, moving out from under region
[...]

PR #1587, opened 2022-08-19
FAIRE-seq assay

hectorguzor:
As discussed in issue #1556 This PR 

1. imports [regulatory_region](http://purl.obolibrary.org/obo/SO_0005836) from SO and makes it the target entity of 'FAIRE-seq assay'. 

2. It makes 'sequencing assay' the asserted parent of 'FAIRE-seq assay'.

3. It  makes [primary structure of DNA macromolecule](http://purl.obolibrary.org/obo/OBI_0000811) the target entity of 'DNA sequencing assay'.


hectorguzor:
The commit history is messy. I will start a new PR with a cleaner commit history.

Issue #1597, opened 2022-09-19
Refresh RO import


cmungall:
It looks like the RO import is old - it still uses the "inheres in" label, we should all be using the latest RO with the "characteristic of" relation

jamesaoverton:
This will be addressed by COB integration work, e.g. WIP #1834.

PR #1618, opened 2022-10-18
add 'cow-baited arthropod trap'

### LINKED ISSUE #1602: NTR: cow-baited arthropod trap
Definition: A resting arthropod trap that consists of a tent inside which an adult cow is tethered to the ground, in order to attract arthropods and collect those that rest on the interior walls of the tent.

Parent: resting arthropod trap

Contributor: Sarah Kelly

johnwjudkins:
Closes #1602.


jamesaoverton:
`external-by-hand.owl` is a bit of a fossil. All that's left in it is imports from RO, which are stale, as mention in https://github.com/obi-ontology/obi/issues/1597. So I think we should replace with a proper import. But that can wait a little longer.

The rest is good. Thanks!

Issue #1621, opened 2022-10-25
Format term-tracker-item annotations consistently


matentzn:
A handful of easy-to-fix cases: https://api.triplydb.com/s/-p00wlgwm

jamesaoverton:
@hectorguzor could you please make these small fixes?

hectorguzor:
@jamesaoverton one of the terms with a bad term-tracker annotation format is an IAO term 
 http://purl.obolibrary.org/obo/IAO_0006011 
 
Below is the term-tracker item for this term. 
https://github.com/information-artifact-ontology/ontology-metadata/issues/40
 
This wouldn’t be a PR for OBI, but IAO to fix right? Should I just create a PR for the IAO repository for this term?

jamesaoverton:
Yeah, please make an IAO PR for that fix. Much appreciated!

PR #1623, opened 2022-10-28
fixes term-tracker format for OBI terms

hectorguzor:
This PR fixes the term-tracker format for OBI terms, as discussed in issue #1621. 

It does not fix the term-tracker format for the IAO term. A separate PR will be created for that in IAO.


jamesaoverton:
@johnwjudkins Can you please do a quick review of this PR, for practise?

Issue #1633, opened 2022-11-17
NTR: organism detection assay (more general than OBI:0001624)


johnwjudkins:
These changes are not urgent.

Label: organism detection assay
Definition: An assay that detects a type of organism and/or the quantity of individuals of a type of organism. 

This is meant to be a more general term than [OBI:0001624](http://purl.obolibrary.org/obo/OBI_0001624) (which is used for NIH metadata, so its meaning should not change). Compared to OBI:0001624, this new assay is not limited to species identification; it can identify organisms by other taxa or life cycle stage, or it can measure density of organisms or count them.

Besides adding the new term, I'd like to replace "identification" with "detection" in the label and definition of OBI:0003064, because I would like this data term to be the output of the new, more general assay, rather than the output of OBI:0001624.

We should also consider relabeling OBI:0001624 to 'organism species identification assay' for clarity.

zhengj2007:
For term OBI:0003064, it should be better to add logical axiom that reflect the textual definition.

johnwjudkins:
Discussed 2022-12-05. May want measurand assay as superclass of analyte assay.

johnwjudkins:
Changes in descendent terms to match this:

Remove the alternative label for 'differential medium assay', because the assay is not limited to bacteria.
All microbiology assay definitions should follow the 'feces microbiology assay' definition with the addition of "organism detection" between "An" and "assay" (so "An organism detection assay that detects microorganisms in a feces specimen." and similarly for other specimen types).
All "organism identification" terms should be "organism detection" except OBI:0001624.

johnwjudkins:
2022-12-12:
'organism identification objective' should be 'organism species detection objective'.
'organism identification assay' can be relabeled as 'organism species detection assay' with equivalent axiom based on 'organism species detection objective'.

Need some note/alt. label noting changes. Alt. label 'organism identification assay' for legacy term.

hectorguzor:
@johnwjudkins it looks like a PR was created and merged for this. Can I close this issue?

johnwjudkins:
@hectorguzor Yes.

PR #1640, opened 2022-12-15
changes to organism detection terms in assay module

johnwjudkins:
This should resolve assay terms related to organism detection mentioned in #1633. (Associated data terms will need to be changed also, but those can be done in a separate request.) See the comment for 'organism detection assay'. I also corrected a misspelling of "reflective" in an unrelated term.


jamesaoverton:
Discussed on the OBI call 2023-01-09: Looks good!

PR #1646, opened 2022-12-27
move 'ANOVA' under 'statistical hypothesis test'

### LINKED ISSUE #1570: ANOVA should be a subclass of statistical hypothesis test to be consistent with STATO
OBI:

image

note that ANOVA and 'statistical hypothesis test' are siblings

STATO:

image

STATO injects its own intermediate "between group comparison statistical test" between two OBI classes. I think injection is problematic, but that is for STATO and OBI to work out.

However, STATO also injects an is-a between two OBI classes, ANOVA and 'statistical hypothesis test' - I am not sure why that was injected into STATO rather than requested directly of OBI

hectorguzor:
This PR resolves #1570. It moves 'ANOVA' as a subclass of 'statistical hypothesis test'.


johnwjudkins (commented):
This may not be a huge issue, but I would edit the textual definition to base it on the new genus. No other issues.

jamesaoverton:
Discussed on the OBI call 2023-01-30: @hectorguzor will update the definition.

johnwjudkins (approved):

PR #1647, opened 2022-12-29
remove universal restriction axiom in 'local field potential recording'

### LINKED ISSUE #1595: remove axiom: local field potential recording has-input that is located *only* in brain
This interesting paper (h/t @jamesaoverton) finds a problem with OBI:

https://www.semantic-web-journal.net/system/files/swj3245.pdf

'local field potential recording':
(has_specified_input some
((tissue and (’located in’ only brain))
and (’has role’ some ’evaluant role’))),

I think the reasons it finds this are interesting (difference between RO and BFO-FOL) but are not relevant for OBI

What it does reveal however, is an anti-pattern:

`located-in only` is almost always unintentional and wrong. We already know that universal restrictions must be avoided with transitive relations like part-of. It's not appreciated that they should also be avoided with certain kinds of role chain structures.

Consider

- tissue1 located-in [a brain]
- [a brain] part-of [a body] (from uberon)
- located-in o part-of => located-in (from ro)

entails:

- tissue1 located-in [a body]

violating the condition

I think this should be uncontroversial that this should be removed, as should any other located-in only axioms

(I am not sure why standard OWL reasoning didn't catch this)

I would go further, my own personal experience is that universal restrictions are almost always useless for reasoning, always cause confusion, and can almost always be rewritten as simpler to understand and computationally and mentally reason with existential axioms, but YMMV.

hectorguzor:
This PR resolves #1595 by removing 'only' from 'local field potential recording' axiom.


cstoeckert (approved):
Change in template is good and checked assay.owl from branch in protege which is also good. Other file differences in template did not change anything.

jamesaoverton:
Discussed on the OBI call: @hectorguzor will resolve the new merge conflict and then we'll merge.

hectorguzor:
I will be closing this PR, and I will create a new one. After attempting to resolve a merge conflict, I made a mistake that resulted in a test fail. I know what the mistake was, but rather than messing around with OBI to fix that, I think it will be better to just create a fresh PR pulling from the latest master. 
Sorry for any inconvenience. I will make sure to reference this PR in the new one.

Issue #1637, opened 2022-12-15
NTR: COA Terms


Daniel-Olson:
This NTR relates to a previous thread regarding Cpath's desire to include COA terms in OBI (https://github.com/obi-ontology/obi/issues/1586). 

This first set of terms were taken from the Friedreich's Ataxia Rating Scale (FARS). Additional information can be found in this template (https://docs.google.com/spreadsheets/d/1LOvA8ILhSKxc5WdcmD0NR-YQyo_WfSB_MqMFUjpuLeM/edit#gid=0) including the desired parent classes.

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/dolson/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/dolson/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</head>

<body link="#0563C1" vlink="#954F72">

ontology ID | label
-- | --
OBI:0003501 | functional assessment
OBI:0003502 | facial atrophy assessment
OBI:0003503 | tongue atrophy assessment
OBI:0003504 | forced cough assessment
OBI:0003505 | speech test assessment
OBI:0003506 | finger to finger assessment
OBI:0003507 | nose to finger assessment
OBI:0003508 | dysmetria assessment
OBI:0003509 | rapid alternating hands assessment
OBI:0003510 | finger taps assessment
OBI:0003511 | heel along shin assessment
OBI:0003512 | heel to shin assessment
OBI:0003513 | muscle atrophy assessment
OBI:0003514 | muscle weakness assessment
OBI:0003515 | vibratory sense assessment
OBI:0003516 | position sense assessment
OBI:0003517 | deep tendon reflex assessment
OBI:0003518 | limits of stability assessment
OBI:0003519 | sitting posture assessment
OBI:0003520 | stance feet apart assessment
OBI:0003521 | stance feet apart assessment, with eyes closed
OBI:0003522 | stance feet together assessment
OBI:0003523 | stance feet together assessment, with eyes closed
[...]

PR #1652, opened 2023-01-24
Adding new COA terms under Assay>physical examination of individual

Daniel-Olson:
Fixes issue #1637 

Supersedes #1641 

@jamesaoverton 

I have updated the definitions per @bpeters42 request and changed the hierarchy to accommodate physical examination of individual.


jamesaoverton:
@Daniel-Olson Can you please replace 'limits of stability test' with its new label 'limits of stability assessment' throughout. That's what's causing the automated tests to fail.

Daniel-Olson:
I added the new OBA IDs to the new template (physical-examination.tsv).

ehartley:
@jamesaoverton We used the OBA IDs instead of the labels in the "assessed biological trait" column because the terms are in the process of being added to OBA as part of [this PR](https://github.com/obophenotype/bio-attribute-ontology/pull/201). So should wait to merge this PR until these new OBA IDs have been released, which will happen soon (Nico is working on it).

jamesaoverton:
Sorry for the series of delays. I made a few changes that you should know about:

1. use the OBA labels in the template, instead of the term IDs
2. change the 'anatomical site' axiom from "'is about' some %" to "has_specified_output some ('is about' some %)"; the domain of 'is about' is 'information content entities', but 'assay' is a 'process', so the old axiom just doesn't work
3. remove some unused columns

PR #1674, opened 2023-03-28
tweak specimen collection device def

ddooley:
Using "device" rather than "container" in def.


mgiglio99 (approved):

mgiglio99:
Reviewed on April 24th OBI call.

PR #1677, opened 2023-03-29
#1512

DanBerrios:
Change nucleic acid extraction (and child processes) to output nucleic acid extract (and corresponding child extracts)


mgiglio99 (approved):
Discussed on April 24th call.

Issue #1634, opened 2022-12-13
term update: genome coverage


cmrn-rhi:
This is a request to update the "genome coverage" [[OBI:0001939](http://purl.obolibrary.org/obo/OBI_0001939)]

**New parent term:** sequence data [[OBI:0000973](http://purl.obolibrary.org/obo/OBI_0000973)]
**New definition:** A sequence data that is the amount of a reference sequence covered by a specific genome of interest, calculated as the total number of generated bases in the sequenced genome divided by the reference/expected genome size.
**Definition source:** https://orcid.org/0000-0002-1107-9135

**Rationale:** Would appreciate a more general definition that isn't eukaryotic centric, with additional information being added as comments / editor notes instead of within the definition.

Originally a new term request as part of issue #1579 ROBOT spreadsheet, but has since been removed from that request.

ddooley:
I just tried editing the obi-edit.owl file with protege 5.6.1 and the text diff is gigantic.  I'm wondering if all OBI editors are now using 5.6.1?  The textual change in the file is often one involving dropping the "string" data type which I guess is a default.:

        <obo:IAO_0000117 rdf:datatype="http://www.w3.org/2001/XMLSchema#string">PERSON:Bjoern Peters</obo:IAO_0000117>
        <obo:IAO_0000117>PERSON:Bjoern Peters</obo:IAO_0000117>

        <rdfs:label rdf:datatype="http://www.w3.org/2001/XMLSchema#string">obsolete_is_described_by</rdfs:label>
        <rdfs:label>obsolete_is_described_by</rdfs:label>

So I can go ahead and do this commit as long as everyone upgraders to protege 5.6.1 .

jamesaoverton:
It's better if I update the ROBOT version and all the OWL files in one PR.

ddooley:
Ok I'll leave that in your hands, and then redo the pull request https://github.com/obi-ontology/obi/pull/1675

jamesaoverton:
#1676 reformatted all the files with the new OWLAPI, so everyone should use the latest Protege and ROBOT going forward.

PR #1679, opened 2023-04-02
genome coverage defn simplification #1634

ddooley:
Line 18376 of foodon-edit.owl diff is where change is.  Unfortunately there's some non-determinism still moving annotations around everywhere.  What to do?


jamesaoverton:
What's the exact version of Protege that you used for this?

ddooley:
image

I don't have to update the Protege OWLAPI separately from installing protege do I?

jamesaoverton:
No, OWLAPI is a fundamental part of Protege, not a plugin.

Do you have other plugins installed?

jamesaoverton:
Damien Goutte-Gattat debugged this in the OBO Slack. I should have run `robot convert` twice for each file in #1676. This should not be a problem going forward.

You can proceed with this PR if you like, or wait for me to update all the files again, which I may not get to for a few days.

ddooley:
Aha!  Ok, go ahead at your leisure.  This pull isn't in a hurry.

PR #1686, opened 2023-04-28
limit of detection terms

### LINKED ISSUE #1485: Add upper/lower limit of detection terms
Not urgent, but 'upper limit of detection' and 'lower limit of detection' are useful terms to add to OBI. They are also interesting examples of value specifications: they are somewhat like settings and predictions, because they have the same "shape" as a measurement datum but mean something very different.

hectorguzor:
close #1485 

This PR creates two limit of detection terms under [value specification](http://purl.obolibrary.org/obo/OBI_0001933).


cstoeckert (commented):
Looks good as discussed.

PR #1694, opened 2023-05-04
relabel donor to donor role

cmungall:
I am not a fan of affixing "role" to labels but consistency is better than perfection

PR #1698, opened 2023-05-10
relabel donor class

hectorguzor:
This PR relabels "donor" to "donor role" and closes #1694.


jamesaoverton:
@cmungall Can you please confirm that this is the only thing you wanted to change in #1694?

PR #1700, opened 2023-05-14
Edits to planned process textual defs

avsculley:
Related to issue #1692


hectorguzor:
Hi @avsculley 

Thank you for this work. Would you be able to create a Google sheet that indicates the changes you are proposing? Something that shows the old definition, new (proposed) definition, and any logical changes that need to be made. That way, we could review these in one of the OBI calls. The OBI group would like to review these before moving forward with the PR. You can link the Google sheet to issue https://github.com/obi-ontology/obi/issues/1692. Thanks again for your work and willingness to contribute.

PR #1706, opened 2023-06-01
Template for 'administration of substance in vivo' terms

sebastianduesing:
Closes issue #1703. Creating a template (administration.tsv) for subclasses of 'administration of substance in vivo' to standardize the design pattern for terms including the route-of-administration terms requested in issue #1703 for IEDB.


sebastianduesing:
I've added a few commits in which I've moved some existing administration-branch terms to the administration template. [Here's a link](https://drive.google.com/file/d/1hLdo_-2DeWUz_kaKSqR3C-glBFa_Jba4/view?usp=sharing) to a ROBOT diff between the existing version (with the current terms in obi-edit) and the proposed version (with those terms in the administration module). I've included some annotations/comments about the changes made; almost all changes to logic have to do with getting rid of redundant class axioms that are already inherited from a parent class. I'm happy to talk through this on an OBI call.

sebastianduesing:
This is ready for final review & merging. To be discussed in next OBI dev call.

PR #1707, opened 2023-06-07
Flow cytometry 1007

### LINKED ISSUE #1007: add terms for flow cytometry gating definitions based on dyes
We need flow cytometry gating terms that  rely on the use of dyes, such as Annexin stains to detect damaged or dying cells that have a ruptured membrane, or CFSE to mark cells that have divided. Terms needed: viable and proliferated.

sebastianduesing:
Closes #1007. Adding terms for dye-based flow cytometry assays as requested by Randi.


jamesaoverton:
Discussed on the OBI developer call 2023-07-10. Ready to merge.

PR #1752, opened 2023-11-28
Obsoleting OBI's CD8 receptor class in favor of PR term

sebastianduesing:
Per #353,  obsoletes OBI's "CD8 receptor" in favor of the PR version (already imported by OBI).


sebastianduesing:
Not sure why this is failing the check—the ROBOT diff looked good.

PR #1762, opened 2024-01-26
Add albumin-free serum

### LINKED ISSUE #493: albumin free serum
Following thread with BIll Hogan working on the analyte assay template in order to allow representation of "ionized calcium assay" and distinguish it from "total calcium assay". The evaluant differs between the 2 assays. 

create a defined material class "albumin free serum" output of material separation, possibly add constraint "lacks albumin"

see thread on IAO

Reported by: proccaserra

Original Ticket: [obi/obi-terms/500](https://sourceforge.net/p/obi/obi-terms/500)

sebastianduesing:
Closes #493. Adds "albumin-free serum."

The diff on this is weirdly big, and I'm not sure why. The ROBOT diff is clean. We can discuss in next OBI call—if there's a reason why this one turned out so messy, I'll re-do this PR.

Editing to add: James said the messy diff is probably OWL API issue, but it looked benign. To discuss on call: the logical axiom "blood serum specimen and (has part only (not albumin))".


cmungall:
I dunno how you ended up with genids, maybe from CLO. If you don't get rid of them they wll continue to cause churn

(of course robot diff by design shows no change as there is no change at the OWL axiom level)

cmungall:
It looks like OBI is interested in modeling media mixtures as might be found in e.g sigma aldritch catalog

Is there a document with the design patterns documented? I think you will encounter issues with OWL quantification further down the line. I have a general interest in this area and in ontological representation of media mixes for microbial culturing, and am planning to use [chemrof:ChemicalMixture](https://w3id.org/chemrof/ChemicalMixture) as DPs for hierarchical quantification of both precise and imprecise mixtures. You can ignore my comment for the purposes of this PR though, just an expression of interest. cc @realmarcin

turbomam:
I did some experimentation with this new class in
- https://github.com/obi-ontology/obi/issues/1767

I never accomplished what I had hoped for, but that doesn't mean I have any objection to this PR

sebastianduesing:
@cmungall I just put together a pull request (#1769) to remove the genids. They were related to OBI:0100061, which had an editor note attached to one axiom; switching that to a general editor note annotation got rid of the genids.

Thanks for the interest in the issue of modeling mixtures. That's a topic that @ddooley and I will likely be working on in our material processing ontology project, though that's still in the early stages. Agreed that OWL quantification will be challenging.

PR #1765, opened 2024-02-07
Add minimum resolved unit of measurement

### LINKED ISSUE #506: resolution/resolving power
a dependent continuant describing the ability of an instrument or an experimental procedure to distinguish between 2 entities of interest.
examples:
In mass spectrometry, resolution is a measure of the ability to distinguish between two peaks of different mass-to-charge ratio, m/z.
In nucleic acid mapping assays, single nucleotide resolution indicates that the assay is able to provide information for individual nucleotide in a nucleic acid molecule

Collaboration with RNAO

Reported by: proccaserra

Original Ticket: [obi/obi-terms/515](https://sourceforge.net/p/obi/obi-terms/515)

sebastianduesing:
Closes #506. Adds "minimum resolved unit of measurement" using Bjoern's proposed label/definition as per comments on that issue, as discussed on OBI dev call.

I'm not sure where the label "minimum resolved unit of measurement" came from, but the research I've done suggests that the term as Bjoern proposed it is probably synonymous with "[least count](https://en.wikipedia.org/wiki/Least_count)." I put "least count" as an alternative term in, but we should discuss whether it would be better to just use "least count" as the label for the term, as that appears to be a more standard/common name for the same thing.


sebastianduesing:
Updated with the change we discussed in the 2024-02-26 OBI dev call: switching "...indicates the smallest _value_ that can be resolved..." to "...indicates the smallest _increment_ that can be resolved..." in the definition. This PR is good to go.

PR #1766, opened 2024-02-07
Add gene silencing assay

sebastianduesing:
This is a draft of a first pass at a term for gene silencing assay. I'm not sure exactly what we would want from a definition or axioms for this term and would appreciate some feedback whenever we talk about this term on the dev call.


sebastianduesing:
Discussed on OBI dev call 2024-02-26. The original issue was actually resolved by OBI:0002628.

PR #1774, opened 2024-03-20
proposed DI water structure

turbomam:
I don't feel confident about how far out this should be axiomitized, or whether deionization should be thought of as a [purification](http://purl.obolibrary.org/obo/OBI_0001505) or as a [material component separation](http://purl.obolibrary.org/obo/OBI_0600014)

I haven't attempted to find a safe range of OBI ids yet.

CC @cmungall


cmungall (changes requested):
- use your ID range
- Should be a subclass of ENVO:water
- Add definitions
- There is an argument for putting in ENVO but as this is primarily a material created artificially OBI is fine

jamesaoverton:
OBI doesn't have user ID ranges. Please use this Google Sheet to get an ID: https://docs.google.com/spreadsheets/d/1tpDrSiO1DlEqkvZjrDSJrMm7OvH9GletljaR-SDeMTI/edit#gid=224833299

We define ENVO imports in this file: https://github.com/obi-ontology/obi/blob/master/src/ontology/OntoFox_inputs/ENVO_input.txt

PR #1777, opened 2024-04-05
Release Candidate 2024-04-05

sebastianduesing:
New terms added:

ID | Label
---|---
OBI:0003652|albumin-free serum
OBI:0003653|fasting
OBI:0003654|overnight fasting
OBI:0003655|fasted
OBI:0003657|minimum resolved unit of measurement


jamesaoverton:
Most of the line differences here are from #1764. I'm a little worried about important terms being lost from OBI Core, so I'll take a closer look when I get time.

jamesaoverton:
Discussed on OBI call 2024-04-22: Sebastian and James will add the 2 IAO object properties and 7 IAO classes that were dropped `obi_core.owl` to the `core.txt` and see if that looks right.

We'll close this PR without committing and make a new release candidate.

PR #1778, opened 2024-04-08
Add editor's note about relation b/t OBI & CHEBI: solvent/solvent role

### LINKED ISSUE #1771: Harmonize OBI:solvent role and CHEBI:solvent
[CHEBI:solvent](https://terminology.nfdi4chem.de/ts/ontologies/chebi/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FCHEBI_46787) seems to be equivalent to [OBI:solvent role](https://terminology.nfdi4chem.de/ts/ontologies/obi/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FOBI_0302732), as it is defined as a role and used in other OBO ontologies as such under BFO:role. Unfortunately, it is not labeled as a role in CHEBI and its textual definition refers to the molecular entity that bears this role not the role itself. So the label and definition including the asserted subclassOf axioms in OBI are more accurate.

The benefit of CHEBI:solvent is that it is being used in CHEBI in subclassOf axioms on those molecular entities that have a solvent role. 

Now I'm wondering is how this should be resolved.
Should the two classes be made equivalent? Should there be an xref annotation?

sebastianduesing:
Closes #1771. Adds an editor's note to OBI:"solvent role" specifying the difference between it and CHEBI:"solvent," using language proposed by @bpeters42.

### DIFF
--- a/src/ontology/obi-edit.owl
+++ b/src/ontology/obi-edit.owl
@@ -26641,6 +26641,7 @@ determinations of compounds.</obo:IAO_0000115>
         <obo:IAO_0000112 xml:lang="en">PMID: 18373502.Transfusion. 2008 Mar 25. Solvent/detergent treatment of platelet concentrates enhances the release of growth factors.</obo:IAO_0000112>
         <obo:IAO_0000114 rdf:resource="http://purl.obolibrary.org/obo/IAO_0000125"/>
         <obo:IAO_0000115 xml:lang="en">solvent role is a role which inheres in a molecular entity capable of ensuring the dissolution of another chemical entity and realized by the process of solvation</obo:IAO_0000115>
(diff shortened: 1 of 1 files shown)


bpeters42 (approved):
reviewed in OBI call 4/15

turbomam:
An editor note like this is a good way of adding clarity, but I beleive that, as of April 2024, roles and material entities are disjoint in ChEBI, and this particular editor note is actually misleading.

2024-04-22 OBI call
It is concerning that roles and material entities in ChEBI may have been less clear in the past, and if a change was made, we don't know about it in this OBI call.

cmungall (changes requested):
I agree with @turbomam. In fact these branches have always been disjoint.

bpeters42:
I cannot reconstruct what I had looked at before, and @turbomam and @cmungall are completely right. We should not add this comment, and go back to the issue.

bpeters42:
Closing this PR to revisit the discussion on the issue itself.

PR #1781, opened 2024-04-22
Add terms to core.txt that were missing after IAO import module update

sebastianduesing:
Adds 2 object properties and 7 classes that were dropped from obi-core after the IAO import module update to `src/ontology/views/core.txt`.


jamesaoverton:
This didn't work out as I expected.

These four terms are no longer in `obi.owl`: 'information carrier', 'denoted by', 'action specification', 'centrally registered identifier registry'. So they're still not in `obi_core.owl`.

I think they need to be added to the IAO import module. @sebastianduesing Can you please do that in this PR?

sebastianduesing:
@jamesaoverton Got it, sorry about that. If I don't get a chance to update the PR with those today, I'll do it on Wednesday morning.

sebastianduesing:
@jamesaoverton Just added those four terms. As far as I can tell, they are now all showing up in `obi_core.owl` as they should.

PR #1790, opened 2024-06-12
Add ring trial study design

### LINKED ISSUE #1768: Add a concept to represent ring trials
A ring trial is an experiment to determine reproducibility by performing the same experiment under different conditions in different laboratories. See for example https://eco-fab.org/about-ecofab-initiative/

I think this is probably best represented as a study design in OBI. I don't think there is a need to create shadow classes under planned process. It may even be best represented as a property of a study.

I can make a PR if someone points me at the best exemplar for this kind of thing.

sebastianduesing:
Closes #1768. Adds "ring trial study design".


DanBerrios:
Discussed, approved.

jamesaoverton (approved):

PR #1792, opened 2024-06-13
Add editor note to solvent role

### LINKED ISSUE #1771: Harmonize OBI:solvent role and CHEBI:solvent

sebastianduesing:
Adds an editor note to `solvent role` using language proposed in that issue discussion.

I'm unsure what needs to be done to map `OBI:solvent role` to `CHEBI:solvent` as suggested in that conversation. If changes need to be made in order to do that, I'll add a commit here.


DanBerrios:
Discussed on OBI Dev call, the issue creator and Bjoern and others should approve closing the issue.

PR #1794, opened 2024-06-18
Making all definitions language tagged

### LINKED ISSUE #1793: Make language tags consistent
Half the OBI definitions have lang=en, the others are plain literals/strings

```
➜  ontology git:(master) ✗ grep IAO_0000115 obi-edit.owl | grep -c lang=
656
➜  ontology git:(master) ✗ grep IAO_0000115 obi-edit.owl | grep -vc lang=
702
```

I don't think there is OBO guidance but I think intra-ontology consistency is always good, it seems your intent is to be explicit, I think it would be clearly safe to auto-convert all definitions to lang=en

dragon-ai-agent:
Fixes #1793


cmungall:
Good point, updated PR for definitions in templates

Do you want a separate PR for other APs?

Does it make sense to do for e.g.

`<obo:IAO_0000119>bp</obo:IAO_0000119>`

jamesaoverton:
The second commit deleted tabs and broke the templates. @sebastianduesing will fix it.

jamesaoverton:
Closing in favour of #1795

PR #1812, opened 2024-08-12
COB Transition Demo 1

sebastianduesing:
COB transition work to be down on 2024-08-12 OBI dev call.


jamesaoverton:
We never, ever remove a term that we've published. Please add OBI:0001720 to the obsolete template.

sebastianduesing:
Sorry, that was a sloppy mistake on my part. I'm not sure how I missed that step. Fixed.

jamesaoverton:
Everything seems to be working except:

1. `COB_based_on` which violated `src/sparql/iri-violation.rq`, see https://github.com/OBOFoundry/COB/issues/280
2. `obi.obo` conversion, because of a problem in COB:

```
OBO STRUCTURE ERROR Ontology does not conform to OBO structure rules:                                                                                                                        multiple comment tags not allowed. in frame:Frame(CHEBI:33252 id( CHEBI:33252)name( atomic nucleus)comment( This is equivalent to CHEBI:33252)comment( Some people may be uncomfortable calling every proton an atomic nucleus)is_a( CHEBI:36342))
```

PR #1814, opened 2024-08-15
Update imports 2024-08-15

sebastianduesing:
Updates all OntoFox imports and makes changes to resolve issues resulting from recent obsoletions, reclassifications, and relabelings. One further commit to replace references to the obsoleted GO term "histone modification" will be added following the pending addition of "modified histone" to PRO (see PROconsortium/PRoteinOntology#336), and this pull request will be finalized following that commit.


sebastianduesing:
At present, this pull request resolves two assays that used the obsoleted GO:'histone modification' term by removing the axiom that used that term. A full fix will be possible after the next PRO release, probably be ~6 months from now, which will include a requested term for '[modified histone](http://purl.obolibrary.org/obo/PR_000085778)' which is live on PRO's site currently but which cannot be imported via Ontofox currently.

In order to proceed with the import refresh without waiting on the PRO release, those two terms, '[histone modification identification by ChIP-chip assay](http://purl.obolibrary.org/obo/OBI_0002016)' and '[histone modification identification by ChIP-Seq assay](http://purl.obolibrary.org/obo/OBI_0002017)', will be left without a replacement for the histone modification axiom for the time being; when this PR is merged, I will open an issue for those two, and as soon as we can import the PRO:'modified histone' term via Ontofox, I'll fix them.

jamesaoverton (approved):

PR #1818, opened 2024-08-29
Fix classes that reasoned into places too high in the hierarchy

sebastianduesing:
Adds subclass axioms to fix a handful of classes that were reasoning into places far too high in the hierarchy, per discussion with James.

James, earlier we talked about whether or not "pool of specimens" should be a subclass of specimen. I said no at the time, but after rereading its definition and the definition of specimen, I've changed my mind. I've put it as a subclass of specimen here. Let me know if you've got any concerns about that.


jamesaoverton:
Related to #1328.

I reviewed and tested this. I don't think it's controversial, and I want it for our OBI-COB work on PR #1812, so I'm going to merge it.

PR #1826, opened 2024-10-03
Added manufactured by Illumina axioms to src/ontology/templates/devices.tsv

turbomam:
Edited as CSV in Libre Office. I don't know why ends of other lines changed but am willing to fix it.

cc @cmungall @aclum


turbomam:
It looks like Libre Office changed some multi-quotation mark strings too

jamesaoverton:
All those quotation marks are a problem. It's a tab-separated file, so it should use normal English quotation rules, and none of the double-double-quotes that comma-separated files require. In Python the setting we use is `QUOTE_NONE`, e.g. https://github.com/obi-ontology/obi/blob/master/src/scripts/sort-templates.py#L12, but that still seems to use some double-double-quotes.

I pushed a manual edit that restores the previous quotation marks, so the only change should be the addition of "Illumina" in the manufacturer column.

Issue #1837, opened 2024-11-18
OBI on COB release process


bpeters42:
Release process suggestion:
- Integrate general ‘clean-up’ changes into master branch
- Release ‘obi-on-the-COB.owl’ (in the views directory; including a URL that points to it)
- Release ‘classic’ obi.owl (with BFO upper level)
- Check if the process to generate the COB version as part of the MAKE file; alternatively have a long lived branch

sebastianduesing:
Completed, see also #1857.

PR #1838, opened 2024-11-22
Pre-COB-transition cleanup

sebastianduesing:
Step 1 of the process described in #1837. Applies some cleanup changes made during the COB transition process:
- Imports the most recent version of OMO instead of non-versioned OMO.
- Relabels "has_specified_input", "has_specified_output", "is_specified_input_of", and "is_specified_output_of" to remove the underscores.
- Renames "alternative term" field to the more up-to-date "alternative label" in all templates.

This PR also includes a minor update to devices.owl to implement a capitalization change made in devices.tsv in a previous PR.


jamesaoverton:
Discussed on the OBI developer call 2024-12-16: I would also include a few more changes from https://github.com/obi-ontology/obi/pull/1834

- drop the useless BFO references
- probably the updates to the IAO import module
- drop external-by-hand if we can
- update the dated RO Core import, and then replace 'inheres in' and 'bearer of' with 'characteristic of' and 'has characteristic'

PR #1839, opened 2024-11-27
Reorganizing 'specimen from organism' hierarchy

### LINKED ISSUE #1833: Organize "specimen from organism" hierarchy by body system
For example, integumentary system specimen, respiratory system specimen, etc.

sebastianduesing:
Closes #1833. Reorganization of the 'specimen from organism' hierarchy in a way that groups specimens by body system. This first pass is a conservative reorganization, in which the only thing I've done is moved existing terms around, e.g., putting 'lower respiratory tract specimen' under 'respiratory system specimen'. This did significantly reduce the massive wall of un-organized specimen terms, but there's still a long, flat hierarchy of specimen terms remaining.

The next step would be to begin creating more grouped-by-system specimen terms, e.g., "digestive system specimen", "circulatory system specimen", "reproductive system specimen", etc. This will require some degree of adoption of Uberon hierarchies. Before I do that, we'll need to come to a decision on how we want to handle that, as is mentioned in the discussion thread of #1833.


sebastianduesing:
Discussed on the 2024-12-16 OBI dev call. Dan pointed out that 'digestive tract specimen' uses subClassOf axioms but should be an equivalent; Sebastian will go through the template and move other terms using subClassOf to be equivalent classes where applicable.

jamesaoverton:
@sebastianduesing Can you please resolve the conflicts with `master`?

sebastianduesing:
Done.

PR #1853, opened 2025-02-27
COB Transition

sebastianduesing:
This PR transitions OBI to using COB as its upper-level ontology. Key changes include the following:

- Replaced BFO import with COB import.
- Replaced RO-core import with a limited import of some RO terms using OntoFox to reduce the number of BFO terms brought in by RO.
- Redirect usages of certain non-COB terms to their COB equivalents, e.g., CHEBI:atom to COB:atom.
- Obsolete some OBI terms that were "donated" to COB, e.g., OBI:planned process, and redirect usages of those terms to their COB equivalents.
- Remove or substitute references to BFO terms that are no longer being imported.


sebastianduesing:
As of 2025-03-03, I just discovered that something is wrong with the processed material entity hierarchy in this PR. Fix coming soon.

turbomam:
@sebastianduesing reviewing processed material hierarchy and will update us next week (March 10th)

sebastianduesing:
False alarm—nothing's actually out of place in the processed material entity hierarchy. The differences I saw were the result of reasoning over the desired changes in the asserted hierarchy. All is as it should be.

PR #1858, opened 2025-03-12
New investigation/clinical trial terms

### LINKED ISSUE #1831: Add term for clinical studies, clinical trials, and observational studies
A question I posed on [Slack](https://obo-communitygroup.slack.com/archives/C04FC8R3Y1K/p1727336334220449): Is there an OBI class that's appropriate for clinical trials? I've found [investigation (OBI:0000066)](https://www.ebi.ac.uk/ols4/ontologies/obi/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FOBI_0000066) which might be a good superclass. Here's what's also in the landscape that I could find on OLS:

1. SIO has a term for clinical trial ([SIO:001000](https://www.ebi.ac.uk/ols4/ontologies/sio/classes/http%253A%252F%252Fsemanticscience.org%252Fresource%252FSIO_001000)), but SIO isn't an OBO ontology
2. OPMI has a term for clinical trial ([OPMI:0004507?](https://www.ebi.ac.uk/ols4/ontologies/opmi/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FOPMI_0004507?lang=en)) but I'm not comfortable reusing terms from this ontology because it seems like it's a pretty use-case-specific ontology, and additionally, it's not using any modern infrastructure so I am not sure I can trust it will be maintained
3. ICO also has a term for clinical trial ([ICO:0000065](https://www.ebi.ac.uk/ols4/ontologies/ico/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FICO_0000065?lang=en)) but I have the same reservations. It looks like there's an overlap of the same authors in these ontologies, so I have reservations to invest in reusing terms they haven't been able to deduplicate themselves.

As someone who wants to annotate clinical trial instances with an appropriate class, I want to have this term in an ontology I can trust :slightly_smiling_face:

Comment from Chris:

> In the past, some ontologies have used this patchwork approach but it leaves us in odd positions where pieces of patchwork get abandoned and no one know where to go to get anything. OBI should claim anything under “investigation” that is biomedical and make OBI IDs.

Comment from James (on slack):

> Yes, OBI 'investigation' includes clinical trials, and it's in OBI scope. We don't tend to subclass 'investigation', for whatever reason, but a good case can be made for a 'clinical trial' subclass. There are various "clinical" 'study design' terms, and many assays.

sebastianduesing:
Closes #1831. This PR adds a handful of investigation terms (housed in a newly-created investigation template, into which I've moved the existing investigation branch) requested in that PR and a few more needed for IAKC. This PR also includes a few additions to `obi-edit`: the terms 'study arm design' and 'study arm population', plus the shortcut relations 'has study design' and 'is study design of', intended to connect investigations to study designs and vice versa. The implementation of these terms is as agreed to on the 2025-03-10 OBI dev call.

See also [this Google Sheet](https://docs.google.com/spreadsheets/d/1OxTLTPFxJLTSP9isPXJ6o8OUitMBtVeMcjNKkDo2MvM/edit?usp=sharing), which includes all terms added in this PR.


cthoyt:
woohoo, thanks @sebastianduesing @jamesaoverton and OBI team!

if you set up a SSSOM template, I can contribute some of the xrefs to other similar/exact terms in other semantic spaces that I tracked down for my presentation

PR #1873, opened 2025-05-20
Add value specifications for HCC

sebastianduesing:
Adds five value specification terms for HCC:

- value not specified
- value other
- value unknown
- value: yes, no, unknown
- Boolean value


jamesaoverton:
I don't love these but it would be good to have something rather than nothing, and I don't have a better suggestion at the moment.

I think that 'value not specified', 'value other', 'value unknown' should be individuals, not classes, but I'm not 100% confident.

The general worry is that these mix epistemology with ontology in a bad way.

PR #1874, opened 2025-05-20
Add assays for HCC

sebastianduesing:
Adds some assays needed for HCC:

- clinical assay
- clinical history assay
- disease severity assay
- genetic characteristic assay


sebastianduesing:
Discussed on 2025-06-02. Sebastian will remove 'clinical assay' as 'physical examination of organism' covers needed terms and we do not want to distinguish all our assays as clinical/non-clinical assays, update 'clinical history assay' to not be specific to humans, and make 'clinical history assay' a child of 'assay', and rename 'clinical history assay' to 'clinical history collection assay'.

PR #1879, opened 2025-06-11
Make 'blood harvesting' a subclass of 'collecting specimen from organism'

### LINKED ISSUE #1878: Term for blood specimen collection
CDE needs a term for the process of blood specimen collection. The OBI term [OBI:1110095](https://www.ebi.ac.uk/ols4/ontologies/obi/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FOBI_1110095) 'blood harvesting' exists, and it seems intended to capture blood specimen collection as one of its axioms is `'has specified output' some 'blood specimen'`, and it is defined as "A material separation where blood is taken from an organism." But it is currently a subclass of 'material component separation' instead of 'collecting specimen from organism', the latter of which should be the parent for CDE's blood specimen collection term.

There are at least two options: (1) make the hypothetical 'blood specimen collection' term a subclass of 'collecting specimen from organism', and keep 'blood harvesting' as a subclass of 'material component separation', in which case its axiomatization should be reconsidered to avoid ambiguity with 'blood specimen collection`, or (2) make 'blood harvesting' a subclass of 'collecting specimen from organism'.

I lean towards option 2. The most obvious reason not to make 'blood harvesting' a subclass of 'collecting specimen from organism' is that there may be cases in which blood is being harvested for objectives other than use as a specimen, e.g., harvesting horseshoe crab blood to extract LAL for bacterial toxin testing. However, 'blood harvesting' is already an ill fit for these cases because of the axiom that states that blood harvesting outputs a blood specimen.

Thoughts? Are there any other reasons not to make 'blood harvesting' a subclass of 'collecting specimen from organism' and create a new 'blood specimen collection' term instead?

sebastianduesing:
Closes #1878. Makes 'blood harvesting' a subclass of 'collecting specimen from organism', to fill a need for a term for blood specimen collection and align 'blood harvesting' with what seems to be intended by its axiomatization and definition.


src/ontology/obi-edit.owl
```diff
@@ -29712,7 +29712,7 @@ MHB 3-5-13: Need to review axiom on this class in light of clarification that it
     <!-- http://purl.obolibrary.org/obo/OBI_1110095 -->
 
     <owl:Class rdf:about="http://purl.obolibrary.org/obo/OBI_1110095">
-        <rdfs:subClassOf rdf:resource="http://purl.obolibrary.org/obo/OBI_0600014"/>
+        <rdfs:subClassOf rdf:resource="http://purl.obolibrary.org/obo/OBI_0600005"/>
```
turbomam, line 29715: shouldn't we change the description too?

The collection of a blood specimen from some organism?

sebastianduesing:
@turbomam Good catch, thanks. I updated it.

I also wonder if we might want to relabel it from 'blood harvesting' to 'blood specimen collection'. Some searching suggests to me that "blood harvesting" is a less commonly-used phrase than "blood specimen collection" for referring to the generalized act of collecting blood from an organism. Search results for "blood harvesting" are mostly about industrial-scale blood collection operations like bleeding horseshoe crabs for LAL or pregnant mares for eCG, with scattered results for the collection of blood stem cells specifically, and a handful of adrenochrome conspiracy theory mentions. Search results for "blood specimen collection" are much more general.

PR #1892, opened 2025-07-02
Add terms for randomized and non-randomized clinical trials

### LINKED ISSUE #1891: Randomized and non-randomized clinical trials
As a follow-up to https://github.com/obi-ontology/obi/pull/1858, I would like to request two new children of [OBI:0003699 (clinical trial)](https://bioregistry.io/OBI:0003699):

1. randomized clinical trial (https://www.cancer.gov/publications/dictionaries/cancer-terms/def/randomized-clinical-trial, https://en.wikipedia.org/wiki/Randomized_controlled_trial)
2. non-randomized clinical trial (https://www.cancer.gov/publications/dictionaries/cancer-terms/def/nonrandomized-clinical-trial)

cthoyt:
Closes #1891

cc @sebastianduesing


sebastianduesing:
Hi @cthoyt, thanks for PRing these terms! We discussed these in the OBI dev call today, and these seem like perfectly reasonable additions. We have a couple small requests:
1. Could you please edit the definition of randomized clinical trial to say "A clinical trial that..." instead of "A study that..."?
2. We'd appreciate it if you could update the OWL file for the investigations module to add these terms to that as well. You can do this by running the command `make src/ontology/modules/investigations.owl` from within the main `obi/` directory.

Once those two things are done, we're happy to proceed with these terms.

sebastianduesing:
Thank you! Looks great.

sebastianduesing (approved):

cthoyt:
@sebastianduesing is there something else that should be done before merge?

PR #1895, opened 2025-07-08
Add relations between investigations, conditions, and interventions (ON HOLD)

### LINKED ISSUE #1890: Add object properties for clinical trials' interventions and conditions
As a follow-up to https://github.com/obi-ontology/obi/pull/1858, it would be great to define two new relations:

1. Clinical trial has intervention
2. Clinical investigation has condition/intervention

These would be able to replace the ad-hoc relations in the following PyOBO code:

https://github.com/biopragmatics/pyobo/blob/b0efcd41c9910194b92a2fdacd3d67d8e3695da9/src/pyobo/sources/clinicaltrials.py#L17-L28

One question is whether these could also be upstreamed into RO?

cc @sebastianduesing

cthoyt:
Closes #1890

In collaboration with @sebastianduesing, this PR adds object properties between investigations and interventions + investigations and conditions. We add the following six object properties for investigations/interventions:

- is intervention in / has intervention
  - is material intervention in / has material intervention
  - is process intervention in / has process intervention

For investigations to conditions, we added two object properties:

- is condition investigated in / investigates condition

We're adding object properties and not annotation properties because these apply to instances of clinical trials. We use [Dasatinib Plus Radiation Therapy/​Temozolomide in Newly-Diagnosed Glioblastoma (clinicaltrials:NCT00895960)](https://clinicaltrials.gov/study/NCT00895960?intr=radiation&rank=1) as an example investigates glioblastoma (DOID:3068) with the following two interventions

- radiation therapy ([MAXO:0000014](https://www.ebi.ac.uk/ols4/ontologies/maxo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FMAXO_0000014?lang=en))
- Dasatinib ([CHEBI:49375](https://www.ebi.ac.uk/chebi/searchI.do?chebiId=CHEBI:49375))

Inside this PR, we add explicit example usage for each added object property.

Side update: in order to make it possible to add structured contributor information via ORCID, I had to update the validate-iri.py script to be aware of the ORCID URI prefix. This should not be controversial.

---

We had originally tried using something like this to connect an investigation to a material entity, but it turns out that there aren't any good ways to have a "macro" like this as part of the definition of the object property

`'has intervention' some ('study intervention' and ('realizes' some ('material to be added role' and ('role of' some [relevant chemical])`


cthoyt:
@sebastianduesing just took a pass at adding more context to all properties and updating the contributor information to include explicit ORCiD for me. You could also swap yours to ORCID if you want. Other than that, I think this is ready for review.

sebastianduesing:
Thank you! Looks good. I'll plan to bring this up for review at next week's OBI dev call.

cthoyt:
just cleaned this up wrt #1903

cthoyt:
@bpeters42 this is the last step before we can have the full output, can you please usher it through and into a new OBI release?

jamesaoverton:
Sorry, we haven't gotten around to this in a long time, and it doesn't look like we'll have resources for it in the foreseeable future.

Before we could go ahead, I need answers to my 3 questions.

I'm going to close this, but it can be reopened if someone will champion it.

PR #1918, opened 2025-10-03
First pass at an OBI workflow script

jamesaoverton:
The idea here is to write a Python package and CLI for working with OBI. It does things that the Makefile is not good at.

The current version:

- reads and writes Ontofox files (used for #1915)
- reads template files, looks for missing terms, looks up missing terms in a table of new terms, and imports new terms found there (used for #1917)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,11 @@
+[project]
+name = "obi"
+version = "0.1.0"
+dependencies = [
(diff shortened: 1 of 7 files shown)


jamesaoverton:
@sebastianduesing and I have reviewed this together. This PR doesn't affect any OBI content, so I'll just merge it.

PR #1935, opened 2025-12-12
Update 'hematology assay'

### LINKED ISSUE #1310: Consider generalizing the definition of 'hematology assay'
The definition of http://purl.obolibrary.org/obo/OBI_0000630 'hematology assay' is:

> An assay that studies blood and blood producing organs using a variety of techniques and instruments

We discussed this on the OBI call on 2021-02-22 and there was concern that this is too narrow. Specifically, we are worried that hematocrit (the ratio of the volume of red blood cells to the total volume of blood) doesn't fit this definition. We propose to generalize the textual definition:

> an assay studying the physiology of blood

jamesaoverton:
Builds on #1934, Resolves #1310


jamesaoverton:
I will use 'bodily fluid specimen' as the evaluant.

PR #1970, opened 2026-04-13
Add new RNA structure mapping assay OBIs for SHAPE

Vicbeg:
Resolves #1964


Vicbeg:
Hi, I have added the 12 new OBI terms to assay.tsv. Could you please take a look and let me know whether everything looks correct? Thank you

sebastianduesing:
Hi @Vicbeg, thanks for creating this PR! We discussed this in the OBI development call today and we have a couple points of feedback. The definitions look great, but we've noticed the information in the "example of usage" field seems more like a description of the chemical involved—could you remove that descriptive portion and keep the PMIDs you listed in that section? We'd also suggest using the unabbreviated names of the chemicals in the definitions instead of the abbreviations. Thank you!

Vicbeg:
Hi @sebastianduesing, thank you for reviewing the changes! I have made the changes you suggest to 'definitions' and 'example of usage' and I have added a reference PMID to each of the rows here. 
I hope everything is correct now!

sebastianduesing:
Thank you so much! Before we can merge this one, I need to make a couple technical tweaks to which files are included in this PR. On your side, do you have the ability to set this PR as editable by maintainers? If you can do that, I'll make those changes in place here, and if not, I'll open another PR based on this one.

Vicbeg:
It should already be set to editable by maintainers. Thanks!

sebastianduesing:
Perfect, thanks! Also, I noticed there's a CHEBI ID in the reagent column for one term, but not the others. If there are CHEBI terms for the reagents for the other terms, could you add those?

Vicbeg:
I checked and that was the only chemical that had a CHEBI ID. I guess a lot of them are new chemicals.

sebastianduesing:
Gotcha, thanks for checking!

sebastianduesing:
Alright, I think this one is good to go. We'll take one last look at it then get it merged within the next couple days. Once it's merged, these terms will show up in OBI after our next release.

Vicbeg:
Thank you. May I ask when you are planning your next release?

sebastianduesing:
We usually make a release about once every two months. Our last one was in mid-March, so probably mid-May. We've had a handful of NTRs completed recently, so I'm aiming for it to be on the earlier side of May.

jamesaoverton:
Discussed on the 2026-05-04 OBI developer call. @sebastianduesing will remove the redundant "measurement datum" from the "output" column and then merge. Thanks!

PR #1974, opened 2026-04-20
Convert IAO import

sebastianduesing:
This PR converts our IAO import module from OntoFox to ROBOT. This one's been on hold pending IAO's adoption of a couple COB classes, which is done as of the most recent IAO release. The updates to term labels (e.g., data item —> data entity) shown in `views/imported_terms/IAO_terms.tsv` reflect other changes appearing in that release.


jamesaoverton:
Reviewed on the OBI dev call today: This is very straightforward so I'll just merge it.

PR #1976, opened 2026-04-22
Reorder a few terms in assay template

sebastianduesing:
This PR moves a few assay terms that were out of numerical order in the template into numerical order.


jamesaoverton:
This should have been handled by 1573d318cd196773770547db0e53d3b5600e6ade.

PR #1983, opened 2026-05-26
Add new terms for bioimaging concepts

### LINKED ISSUE #1894: Bioimaging concepts
I want to describe bioimaging assays. Currently there is the [imaging assay](http://purl.obolibrary.org/obo/OBI_0000185) subtree and the [Image](http://purl.obolibrary.org/obo/IAO_0000101) entity. However, there are some other concepts that I don't think can easily be described, so I would like to make the following proposals

* Add an `Axis` entity, which I guess would go under [information content entity](http://purl.obolibrary.org/obo/IAO_0000030). Each image has one or more Axes. These are commonly XYZTC where T is a timepoint and C is a Channel.
* Add a Channel entity (also an information content entity), which describes one of the element of the channel `Axis` (note: it is unlikely that we will want to describe each XYZ coordinate or each timepoint). Channel could be imported from [here](https://w3id.org/reproduceme#Channel) or [here](http://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Channel).
* We add a [`Fluorophore`](http://purl.obolibrary.org/obo/MI_0856) entity as a subclass of [Material Entity](http://purl.obolibrary.org/obo/BFO_0000040) (you could also model this as a role played *by* a Material Entity).
* Each Channel might capture one [fluorophore](http://purl.obolibrary.org/obo/MI_0856), so we need a property to link these if one doesn't exist.
* Image Processing such as deskewing, devolution, cropping etc should be captured as new subclasses of [data transformation](http://purl.obolibrary.org/obo/OBI_0200000).

mcmero:
Closes https://github.com/obi-ontology/obi/issues/1894.

This PR adds several new terms related related to bioimaging:

- channel entity
- image transformation, with subclasses that represent common image operations:
    - image segmentation
    - geometric image transformation
    - point operation

### DIFF
--- a/src/ontology/obi-edit.owl
+++ b/src/ontology/obi-edit.owl
@@ -30416,6 +30416,113 @@ realizes some &apos;host of immune response role&apos;</obo:IAO_0000232>
     
 
 
+    <!-- http://purl.obolibrary.org/obo/OBI_0003916 -->
+
+    <owl:Class rdf:about="http://purl.obolibrary.org/obo/OBI_0003916">
+        <rdfs:subClassOf rdf:resource="http://purl.obolibrary.org/obo/IAO_0000027"/>
+        <obo:IAO_0000111>channel</obo:IAO_0000111>
+        <obo:IAO_0000112>DAPI fluorescence channel in a confocal microscopy image.</obo:IAO_0000112>
+        <obo:IAO_0000115 xml:lang="en">A data item that is a component of a digital image that represents spatially registered measurements collected for a specific spectral band, detector, or sensor output.</obo:IAO_0000115>
+        <obo:IAO_0000117>Lachlan Whitehead</obo:IAO_0000117>
+        <obo:IAO_0000117>Pradeep Rajasekhar</obo:IAO_0000117>
+        <obo:IAO_0000117>Marek Cmero</obo:IAO_0000117>
+        <obo:IAO_0000117>Michael Milton</obo:IAO_0000117>
+        <obo:IAO_0000118>image channel</obo:IAO_0000118>
+        <obo:IAO_0000118>fluorescence channel</obo:IAO_0000118>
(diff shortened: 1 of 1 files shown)


sebastianduesing:
Hi @mcmero, thanks for creating this PR! We reviewed this on the OBI dev call today (2026-06-08), and we're happy to add these terms, but we'd like to request a couple small changes. Firstly, the label 'channel' is too broad for what the term is supposed to encompass; we propose changing the label to 'digital image channel' instead. Secondly, we'd appreciate if you could add example of usage annotations with links to publications that make use of relevant image transformations on the image transformation terms. Thank you!

mcmero:
Thanks for reviewing @sebastianduesing and OBI team. Good suggestion to make channel more specific, I've updated this in the PR.

Just to clarify with the second point, do we need an example for image transformation itself or only the three subclasses? Also, the definition source papers/textbooks already contain some of these examples, is it a matter of explicitly linking them here? And if our example is 'Rotation, translation, scaling etc.' are you looking for an example publication that uses each of these techniques?

sebastianduesing:
Sorry for the delay in responding, and thanks for making that update. I think the group wanted examples for the three subclasses of image transformation, not for image transformation itself. One example publication for each subclass would be great—you don't need to provide one for each item in the existing list of examples. If the definition source papers contain those examples, feel free to just copy links to those papers into the example of usage field. All we're really looking for is some publication that shows that these image transformation processes are used in some biomedical investigation(s).

mcmero:
Thanks for the clarification @sebastianduesing. I've now added some citations to the example of usage field for those terms.

jamesaoverton:
Discussed on the OBI call 2026-06-29: Thanks for the changes, we will merge this.

jamesaoverton:
This isn't merging cleanly, so I created a new PR #1988 with exactly the same content, and I'll merge that instead. Thanks again!

PR #1995, opened 2026-06-29
Convert ENVO, OMIABIS, and OMRSE to robot imports

sebastianduesing:
Batch conversion of three small import files. There's no diff in `views/imported_terms/*_terms.tsv` for any of these because the conversion did not affect which terms were included in these files.


src/ontology/robot_outputs/OMIABIS_imports.owl
```diff
@@ -22,49 +22,23 @@
...
-        <rdfs:label xml:lang="en">imported from</rdfs:label>
-    </owl:AnnotationProperty>
-    
-
-
     <!-- http://www.w3.org/2000/01/rdf-schema#label -->
 
-    <owl:AnnotationProperty rdf:about="http://www.w3.org/2000/01/rdf-schema#label">
-        <obo:IAO_0000111>label</obo:IAO_0000111>
+    <rdf:Description rdf:about="http://www.w3.org/2000/01/rdf-schema#label">
```
jamesaoverton, line 39: Why did this change to `rdf:Description`?

sebastianduesing:
While trying to figure out what caused the change to rdf:Description in the OMIABIS file, I learned that OMIABIS has been deprecated. The 2 terms we were importing from OMIABIS are now in OBIB. Should I replace the OMIABIS import with an import of OBIB?

jamesaoverton:
@sebastianduesing Hmm. That's the kind of thing I'd like to seen in a separate PR.

sebastianduesing:
Closing this because we decided to replace the OMIABIS import with an OBIB import (#1997). I'll create a new PR for the ENVO & OMRSE conversions.

PR #1997, opened 2026-07-01
Use OBIB instead of OMIABIS

sebastianduesing:
OBI currently imports two terms from OMIABIS, an ontology which has been deprecated. OBIB has created replacements for those terms. This PR imports OBIB in place of OMIABIS and rewires usages of the OMIABIS terms to their OBIB equivalents. You can see by comparing `views/imported_terms/OBIB_terms.tsv` and `views/imported_terms/OMIABIS_terms.tsv` to see that we would be importing the same terms from OBIB that we were getting from OMIABIS.

If we go through with this PR, I'll update #1995 to remove the OMIABIS conversion files.


sebastianduesing:
Discussed on the 2026-07-13 development call; agreed to make this change.

PR #2007, opened 2026-08-10
Fix "microbioal" typo in three assay definitions

### LINKED ISSUE #2006: "microbioal" typo in three microbiome assay definitions
`src/ontology/templates/assays.tsv` spells "microbial" as "microbioal" in three definitions:

| line | term | label |
|---|---|---|
| 384 | OBI:0002760 | metatranscriptome profiling assay |
| 387 | OBI:0002763 | 16s ribosomal gene sequencing assay |
| 389 | OBI:0002765 | microbiome protein expression profiling assay |

turbomam:
Fixes 
- https://github.com/obi-ontology/obi/issues/2006


sebastianduesing:
Thanks for catching this typo, Mark. Could you please run `make src/ontology/modules/assays.owl` to propagate this edit to the assay module's OWL file and commit that file as well?

turbomam:
@sebastianduesing and @jamesaoverton: I rebuilt `assays.owl` and pushed it, but the CI build fails on the first module because robot can't retrieve this dependency:

https://github.com/obi-ontology/obi/blob/b29b902bc7f2a7b154d96185b3ae2865e9a9de04/src/ontology/obi-edit.owl#L16

```owl
    <owl:Ontology rdf:about="http://purl.obolibrary.org/obo/obi/dev/obi-edit.owl">
        <owl:imports rdf:resource="https://protege.stanford.edu/plugins/owl/dc/protege-dc.owl"/>
```

I can load that `protege-dc.owl` in a browser and with curl, but Java clients (robot, the OWL API, Protégé) refuse the connection. My research suggests that the server sends only its leaf certificate, with none of the Let's Encrypt chain above it. That chain now runs through a new root (ISRG Root YE) that Java's truststore doesn't have. But I dont' fully understand that myself.

```shell
echo | openssl s_client -connect protege.stanford.edu:443 -servername protege.stanford.edu 2>/dev/null |\
grep -E 'Certificate chain|^ [0-9] s:|Verify return'
```

> Certificate chain
>  0 s:CN=protege.stanford.edu
> Verify return code: 21 (unable to verify the first certificate)

This PR passed CI on Aug 10 and nothing else in the repo has run CI since, so any branch will fail the same way until Stanford serves the full chain or OBI stops fetching that import live. I'm looking for the right place to report it to the Protégé folks.

turbomam:
I got the module built locally by adding the missing Let's Encrypt intermediate to a copy of Java's truststore. I'm looking into where I should raise this concern upstream.

turbomam:
I raised this in the OBO community Slack's general channel and replies are coming in.

